Issue #16, opened 2023-07-22
BE Gens excluded in Atlite are Not added to network


ktehranchi:
```
        #TODO: Gens excluded from atlite profiles bc of landuse/etc will not be able to be attached if in the breakthrough network. The atlite tool would create extendable carriers for all busses in the network, but we exclude certain buses if they are removed for landuse/protected area purposes. The BE network has plants that do not have cooresponding atlite bus profiles because those regions were excluded in the rule "build_renewable_profiles".

        One fix would be to generate profiles for all buses, regardless of the land use availability. The land use availability would limit the p_nom_max for expansion, but if we needed to assign solar at those nodes, we could still do so.

        Another option is to revisit the location of BE nodes... need to investigate why those buses are the ones not having profiles built for them.
```

https://github.com/ktehranchi/pypsa-breakthroughenergy-usa/blob/a2dadd26175709acbac845cfd0963d0846697aae/workflow/scripts/add_electricity.py#L759

PR #25, opened 2023-07-26
Renewable Potential Fixes

ktehranchi:
Fixed Renewable potential calculation stemming from issue of CRS in land area calculation. This partially solved the issue of a few solar and wind plants not being added to the network #16 . But will need to address that issue in the future


ktehranchi:
This PR involved editing the pypsa-eur script for build_renewable_profiles and thus removed it from subworkflow. I could keep it as a subworkflow if I PR to pypsa-eur to turn the CRS into a parameter

PR #41, opened 2023-09-08
Adds states to buses

trevorb1:
Included in this PR is: 
- Refactor of `build_base_network.py`
- Refactor of `build_shapes.py`
- Corrected countries being mapped as BAs for each bus 
- Add in states being mapped to each bus


ktehranchi (commented):
Looks good!

ktehranchi:
@trevorb1  one note about the "countries" designation: I kept the countries field for each bus since the clustering scripts we use from pypsa-eur and Pypsa-earth cluster based on the country field.... 

I like the changes you made here, but I just want to call out how in the simplify_network script I retain the country field in order to integrate with the other workflows. 

When we integrate with Pypsa-earth  down the road, I would PR the change from country to "zone" or some other region name

ktehranchi:
@trevorb1 FYI I just made changes to retain state boundaries through the clustering rule in another branch. Will PR over the weekend/monday. No need to work on that.

PR #44, opened 2023-09-14
Corrects cost data for USA 

trevorb1:
In this PR I have: 

- [x] Added logic to retrieve ATB, ATB Transportation, EIA, and PyPSA-Eur costs data 
- [x] Integrated ATB data into the workflow via the `build_costs.py`
- [x] Added regional multiplier data for capital costs from [this](https://www.eia.gov/analysis/studies/powerplants/capitalcost/pdf/capital_cost_AEO2020.pdf) EIA report
- [x] Applied the regional cost multipliers for capex 
- [x] Applied the regional cost multipliers for fuel costs 
- [x] Applied the temporal cost multipliers for fuel costs


trevorb1:
**DONT MERGE YET**

While all the logic is there to add in the majority of capital and fuel costs, there is a couple implementation questions that need to be addressed first I think. 
1. The aggregation of `CCGT` and `OCGT` into one gas `technology`. Right now, the code will average these costs, but not sure if this is the best way. Do all data sets (Breakthough, PyPSA-USA, ADS) track OCGT and CCGT separately? 
2. Do you think doing a comparison between PyPSA-Eur and these updated costs is worth it, just to track if there are any major differences?

ktehranchi:
> **DONT MERGE YET**
> 
> While all the logic is there to add in the majority of capital and fuel costs, there is a couple implementation questions that need to be addressed first I think.
> 
> 1. The aggregation of `CCGT` and `OCGT` into one gas `technology`. Right now, the code will average these costs, but not sure if this is the best way. Do all data sets (Breakthough, PyPSA-USA, ADS) track OCGT and CCGT separately?
> 2. The capex cost multipliers are applied to the annualized investment cost, rather than the true capital cost due to the existing implementation in the `load_costs()` function
> 3. Do you think doing a comparison between PyPSA-Eur and these updated costs is worth it, just to track if there are any major differences?

RE: 

1 Yes, we definitely want to separate those.... This would be handled in the `constants.py' file and we would need to follow suit in our configuration files. It seems like it makes sense to use the `carrier` attribute to track fuel type usage for sector coupling, and the `type` attribute of generators to track the technology type. But when you look at the clustering rule and implementation in pypsa-eur they choose to stick with technology types in the `carrier`.... Perhaps short term we assign technology types in the `carrier` field?

2: Hmm we shouldn't apply the capex cost multipliers to the OPEX.... Options are to change load costs or to apply them to the ``tech_costs` table before being read into the file? I think either would work.

3: Yep probably worth an eyeball to make sure ball-parks are the same... but not worth too much time since we know they'll be different.

trevorb1:
1. Agreed! Maybe I will hold off on trying to implement this until after this code is merged, then create a new issue for separating CCGT and OCGT. 
2. Again, agreed. The reason I didn't do this is because the capital cost is calculated in the `load_costs()` function, as you mentioned. So I would have to pass around the multiplier data throughout the workflow, as `load_costs()` is called downstream of `add_electricity`. I didn't really think of just applying it to the `tech_costs` table before reading in, and will do that. Thank! 
3. Yeah, I agree! I will try to do a quick look to see how far off we are.

trevorb1:
Oh, actually on second thought for the cost multiplier. I don't think we can apply it to the `tech_table` costs before being read in because it needs to be broken out at a state level. :(

ktehranchi:
Okay- so we'll just pass it through load costs. This is where this stuff gets a little tricky when we think about integration with pypsa-eur or earth. I imagine we want to be able to align with their rules as much as possible, but once we change one common helper function then we deviate from their workflow.... I think we should pin down integration with other workflows sooner than later. BC then we risk building up a long re-integration task for ourselves down the road.

trevorb1:
On point 3, here is the comparison table for cost differences between pypsa-eur and pypsa-usa that are greater then 5% off (excluding battery storage costs, as they are way way off and will be a separate issue to investigate). A couple crazy ones near the ends, but I guess all within the same ballpark 

| technology                     | parameter     | unit      | value_eur | value_usa | difference |
|--------------------------------|---------------|-----------|-----------|-----------|------------|
| SMR                            | FOM           | %/year    | 5         | 1.407     | -71.86     |
| solar-utility                  | lifetime      | years     | 40        | 20        | -50        |
| solar-rooftop commercial       | lifetime      | years     | 40        | 20        | -50        |
| solar-rooftop                  | lifetime      | years     | 40        | 20        | -50        |
| biomass                        | efficiency    | per unit  | 0.47      | 0.253     | -46.1702   |
| nuclear                        | VOM           | USD/MWh_e | 3.745     | 2.47      | -34.0454   |
| solar-rooftop commercial       | FOM           | %/year    | 1.57      | 1.052     | -32.9936   |
| solar-utility                  | FOM           | %/year    | 2.48      | 1.732     | -30.1613   |
| central solar thermal          | FOM           | %/year    | 1.4       | 1.021     | -27.0714   |
| CCGT                           | FOM           | %/year    | 3.35      | 2.471     | -26.2388   |
| coal                           | lifetime      | years     | 40        | 30        | -25        |
| geothermal                     | lifetime      | years     | 40        | 30        | -25        |
| geothermal                     | FOM           | %/year    | 2.36      | 1.808     | -23.3898   |
| solar-rooftop                  | FOM           | %/year    | 1.42      | 1.097     | -22.7465   |
| biomass                        | FOM           | %/year    | 4.53      | 3.502     | -22.6932   |
| home battery storage           | lifetime      | years     | 25        | 20        | -20        |
| battery storage                | lifetime      | years     | 25        | 20        | -20        |
| Pumped-Storage-Hydro-bicharger | FOM           | %/year    | 1         | 0.829     | -17.1      |
| ror                            | FOM           | %/year    | 2         | 1.69      | -15.5      |
| OCGT                           | efficiency    | per unit  | 0.41      | 0.351     | -14.3902   |
| CCGT                           | efficiency    | per unit  | 0.58      | 0.547     | -5.68966   |
| OCGT                           | lifetime      | years     | 25        | 30        | 20         |
| CCGT                           | lifetime      | years     | 25        | 30        | 20         |
| solar-rooftop                  | discount rate | per unit  | 0.04      | 0.048     | 20         |
| OCGT                           | FOM           | %/year    | 1.78      | 2.182     | 22.58427   |
| hydro                          | lifetime      | years     | 80        | 100       | 25         |
| ror                            | lifetime      | years     | 80        | 100       | 25         |
| coal                           | FOM           | %/year    | 1.6       | 2.118     | 32.375     |
| offwind                        | FOM           | %/year    | 2.32      | 3.154     | 35.94828   |
| nuclear                        | FOM           | %/year    | 1.4       | 1.968     | 40.57143   |
| nuclear                        | lifetime      | years     | 40        | 60        | 50         |
| biomass                        | lifetime      | years     | 30        | 45        | 50         |
| central solar thermal          | lifetime      | years     | 20        | 30        | 50         |
| Pumped-Storage-Hydro-bicharger | lifetime      | years     | 60        | 100       | 66.66667   |
| onwind                         | FOM           | %/year    | 1.22      | 2.348     | 92.45902   |
| SMR                            | lifetime      | years     | 30        | 60        | 100        |
| hydro                          | FOM           | %/year    | 1         | 2.826     | 182.6      |
| coal                           | VOM           | USD/MWh_e | 3.745     | 13.62     | 263.6849

trevorb1:
Also, for future reference, here is the code I used to replicate these results 

```python 
# get this value from constants.py
EUR_2_USD = 1.07

eur_costs = pd.read_csv("../workflow/resources/costs/2030/pypsa_eur.csv")
usa_costs = pd.read_csv("../workflow/data/costs_2030.csv")

eur = eur_costs.copy()
eur.loc[eur.unit.str.contains("EUR/"), "value"] *= 1.07
eur.unit = eur.unit.str.replace("EUR/", "USD/")
eur = eur.set_index(["technology", "parameter", "unit"])

usa = usa_costs.set_index(["technology", "parameter", "unit"])

compare = eur.join(usa, on=["technology", "parameter", "unit"], lsuffix="_eur", rsuffix="_usa")
compare = compare[["value_eur", "value_usa"]]
compare["difference"] = (compare["value_usa"] - compare["value_eur"]) / compare["value_eur"] * 100
compare = compare.reset_index()
compare = compare[(compare.difference > 5) | (compare.difference < -5)]
compare = compare.sort_values(by="difference")
```

ktehranchi:
Yep, some of these look too far off. Coal VOM looks a bit high compared with these
- [CAISO report](http://www.caiso.com/documents/variableoperationsandmaintenancecostreport-dec212018.pdf) (page 11) 
- 6.925 $/mwh [2021 NREL ATB dashboard](https://atb.nrel.gov/electricity/2021/definitions#operationsexpenditures)
- Also screenshot of the 2023 ATB for New Coal shows VOM costs @ 8.46 $/MWH
Screen Shot 2023-09-22 at 10 01 09 AM


What was the code you used to match with the ATB ?

trevorb1:
oops, should have mentioned this is 2030 costs! Sorry! 

For most technologies, there was a representative value I used, however, for a few (like coal) there isn't. In these cases I followed whatever technology was marked as default, as per [this quote](https://atb.nrel.gov/electricity/2023/other_technologies_(eia)) on NREL ATB 

> The default technology detail best aligns with recent or anticipated near-term installations. Scenarios are labelled with ATB names but correspond to AEO scenarios.

Following that, I landed on the coal VOM 2030 cost as $13.62/MWh

image

You do bring up a good point though, on what is really meant by `Coal-New` then. haha. Maybe just coal without CCS. 

I guess the options are to: 
1. Change the defaults to closer align to PyPSA-Eur (for all techs)
2. Leave as is and document what we are using as default. 

To be implemented is a way to select what tech-detail you want from the NREL ATB. The functions are set up to handle this already, just the logic needs to be implemented to capture this information from the config and pass it into the functions. This will come in a separate PR though

ktehranchi:
Ohh Okay, I think we should try to align with ATB candidate technology type conventions since that's what is used as the standard for US modeling.... But it might be an over-kill to include every single ATB tech-detail variation?

What if we included both coal-new and coal-ccs as separate tech types/ carriers?

This is where it would be nice to properly label `generator.carrier`  as coal, and `generator.type` as the tech type...

ktehranchi:
Yep that sounds good!

trevorb1:
Just an update on this, as I was hoping to merge yesterday! I feel like all the logic is there, but I keep getting a negative objective cost when I run the model. Ive traced the error back to the capital costs, but not sure what is causing the error. I will sort it and merge soon! :)

trevorb1:
Okay, this one was kinda odd. haha. A couple issues arose from trying to sort out this issue though: 
1. On the **master branch** right now (for at least the following network configuration), capital costs are not being assigned to some extendable generators. The code snippet below is looking at the solved network. 
```yaml
network_configuration: "ads2032" # "pypsa-usa" or "ads2032" or "breakthrough" 
```
```python 
n.generators[(n.generators.p_nom_extendable == True) & (n.generators.capital_cost == 0)]
```
image

When this PR applies the regional multipliers, the capital costs are added to generators. But they should also probably be applied when all the generators are added? Because (for example), there are no regional multipliers for oil, so their capital costs are still zero. If oil is turned on to be extendable, this could be an issue. 

2. The issue of negative objective costs is (I think) traced back to only generators that have `p_nom_extendable` as `True`. Looking at the results, some generators will have `p_nom_opt` as a value less than the `p_nom` value. If I fix the `p_nom_min` to be the same as `p_nom` (rather than the default of zero), the objective cost will be positive. I am guessing that when `p_nom_min = 0`, the model will sell off the asset if not needed (only a guess though!!). I believe we were not seeing this behaviour with the model on the master branch due to the issue in point 1, that capital costs are zero for extendable generators. I think these are the relevant constraint equations.

[Objective function term](https://pypsa.readthedocs.io/en/latest/optimal_power_flow.html#objective-function)
image

[Generator power term ](https://pypsa.readthedocs.io/en/latest/optimal_power_flow.html#generator-constraints)
image

I have added in a temporary fix to the bottom of `add_electricity` to set all `p_nom_min` to be  equal to `p_nom` for extendable generators. This will result in a positive objective cost. 

@ktehranchi; On point 1, I will probably just turn that into a new ticket once this is merged? On point 2, are you okay with this temporary fix, and I will add a new issue ticket once merged? Or have you run into a similar issue before and I am missing an easy solution? It is totally possible I am way off base here! haha. 🙃Thanks so much!!

ktehranchi:
1: Thanks for finding this issue. I hadn't assigned capital costs to (I think) any generators in the ads2032 configuration since I had intended for this config to be used mainly for operations-only simulations (for transmission planning case). Moving forward what do you think if we focus on the pypsa-usa configuration to be our default config since that'll cover the entire US. When we get to the point that this is a multi-period model then we can think about folding in the ads2032 into the default config.

I can add some warning settings to let people know they should throw the ads2032 into operations simulation-only mode.... I have a PR coming soon that has an operations-only config setting.


2: Sounds good. Setting `p_nom_min == p_nom` checks out, but I think it would be nice to know retirement decisions in addition to new build decisions. I'm not sure how existing plants retirement CAPEX is treated.... I imagine those assets could be sold off (??). Let's keep an issue ticket open and revisit later.

Let's merge dis-ish! Nice work!

trevorb1:
Great, thanks! 

> I hadn't assigned capital costs to (I think) any generators in the ads2032 configuration since I had intended for this config to be used mainly for operations-only simulations (for transmission planning case).

Ahhh. I see, that makes sense. Maybe we pull all cost assignments out of when generators are assigned, and instead create a new function to assign CAPEX. Then if we are running a operation-study (ie. ads2032), we can just skip over it. I agree though, and I will add this as a new issue. 

> Moving forward what do you think if we focus on the pypsa-usa configuration to be our default config since that'll cover the entire US.

Agreed! I updated the PR to reflect this change. This does lead into possibly a bigger issue, of how to test all these different networks. But that feels like a question for down the road! :)

ktehranchi:
RE: Capex assignment. I was just thinking about this more- We need to be able to capture plant retirements, and the "true" amount of capacity necessary for future systems. I think what would make sense is to have `0 <= p_nom_max == p_nom` for existing generators, with their capital costs == 0. This way the old plants could retire at no cost. Then we would create dummy new generators of each type at each node where `p_nom ==0`, capital cost == true capital cost, ` 0 <= p_nom_max <= 999999`. We would add those extra dummy plants in the `add_extra_components` rule.

Also we should add the build_year per EIA860 data, and have the plants p_nom automatically set to 0 after their lifetime.

Thoughts? Does this check out to you? We can do this in a future PR, but throwing it out now.

Issue #79, opened 2023-09-30
Move Zenodo links out of the `config.yaml`


trevorb1:
### Feature Request

More of a question, but having the Zenodo links in the config file seems odd. Since the downstream logic is set up to only work with the data structures returned from these specific Zenodo deposits, it seems like these should not be configuration options. Rather, they should be hard coded into the `Snakefile`? 

https://github.com/PyPSA/pypsa-usa/blob/bb9e5129e03e952cadc9a9f6e3d7cdb5914ebb88/workflow/config/config.yaml#L20-L26

@ktehranchi, what are your thoughts on this?

ktehranchi:
Yep- we should do this! I can get this done

ktehranchi:
Solved with https://github.com/PyPSA/pypsa-usa/commit/9a75ef918eebd2c791b532f0f03ea18696db8ee6

ktehranchi:
closed by #109

Issue #70, opened 2023-09-26
Integrate Texas and Eastern Interconnects


ktehranchi:
### Feature Request

This is a long-time standing back-burner task to integrate Texas and Eastern interconnections. Currently, only WECC runs in pypsa-usa configuration.

### Suggested Solution

- Map load data balancing authorities to the BA names in the network. This could mean aggregating some of the given BA's in the south. 
- Fixing BA shape files to fill gaps between BAs where nodes may lay
- Build a new cutout for the entire US or multiple cutouts for each interconnection

trevorb1:
@ktehranchi, how far into this issue are you? No rush, but just note in my fix for issue #76 (see PR #84) I have created new cutouts for each interconnection. I can pass those along if needed. Moreover, I am happy to help with the other items on this list as well if needed :)

ktehranchi:
I haven't started on this, but this is in my docket for this week. Thanks for making these build cutout changes... Did you build the cutouts locally already? Whats your thoughts on have the cutouts pre-built and loaded on zenodo versus using the build_cutouts?

If we were to go the zenodo route, it would make sense to just have one cutout for the entire US for each year 2017-23, then filtering the cutout spatially for each interconnect... otherwise we would have 4 cutouts X 6 years.

I can complete the tasks for the load and plant data + GIS work on this issue.

trevorb1:
I created the cutouts for `Western`, `Eastern`, and `Texas` locally (for 2019), and the one for `usa` is currently running. 

I totally agree with your point on keeping them on Zenodo, and the `build_cutout` rule only being triggered if the user specifies it in the config file. Moreover, I like your point of to"just have one cutout for the entire US for each year 2017-23". Probably easier than having different interconnect cutouts!

If we are wanting to give users the option to create a cutout for their interconnection though (through `build_cutout`), we will probably need to change the `cutout` argument to be an input function for any rule that uses it, I think. Something like (untested) 

```python 
def cutout_file(wildcards):
    if config["enable"].get("build_cutout", False):
        return "cutouts/" + CDIR + "wildcards.cutout.nc"
    else:
        return "cutouts/" + CDIR + "wildcards.interconnect_wildcards.cutout.nc"
```

ktehranchi:
Cutouts now uploaded to sandbox [zenodo](https://sandbox.zenodo.org/record/1245562) with texas and eastern interconnect for 2019

trevorb1:
[...]

Issue #64, opened 2023-09-25
Drop buses not mapped to regions 


trevorb1:
### Feature Request

During the process of building the network, each bus is mapped to a `balancing_area`, `state`, and `country`. However there are a subset of buses that are not mapped to any or only a subset of categories. The screenshot below is showing the pre-clustered network. 

![image](https://github.com/PyPSA/pypsa-usa/assets/67297083/db846b57-e2e3-470e-b7f6-d2bbbd6b79d5)

It would be good to either fix the mapping (for those buses that have a BA, but no state/country), or drop the unused ones. Or if the categorized buses are still used, to fix the mapping. Especially from a new user perspective, this can be quite confusing. 

### Additional Info

I believe in a conversation with @ktehranchi, the buses which are not mapped to a region are offshore wind farm location from the breakthrough energy dataset. But it would be good to clarify this

PR #89, opened 2023-10-06
#70 Integrate Texas Interconnect

### LINKED ISSUE #88: Add Offshore Buses for Texas
### Feature Request

The BE & TAMU network has offshore buses for WECC and Eastern Interconnect but not texas. 

### Suggested Solution

Suggestion: Remove all offshore buses in the entire network and add offshore buses using the same method for the entire US. One way to do this would create a meshed grid overlay of the offshore shapes, assigning a bus to the center of each grid point, then filter out the busses using a buffer distance from shore. 

Each remaining bus point would get a voronoi cell, and we could also explore meshed network configurations using this method.

### Additional Info

_No response_

ktehranchi:
Closes #88 #70 #64 #79 

- [x] Finishes Integration of Texas Interconnect
- [ ] Integrates Eastern Interconnection - Some additional GIS Clean-Up is needed to finish integrating eastern interconnect.
- [x] Adds Exclusive Economic Zones as Offshore Shapes for entire US
- [x] Builds cutouts for Texas and Eastern Regions 
- [x] Removes old Offshore Buses, and assigns new for entire US 
- [x] Moves zenodo links out of config, #79 

## Checklist

- [ ] A release note `doc/release_notes.rst` is added.


ktehranchi:
@trevorb1 I've finished integrating texas and reworking how offshore wind buses are added to the network. There is still some work to be done on cleaning up the eastern interconnection GIS shapes, but am hoping to merge this PR first so a researcher from Rice University (Chen Chen) can start working with texas sooner.

I've tested a few different configurations on my end for TX and WECC and they run smoothly. 

One question I have is in the handling of Offshore buses during clustering. Currently I have the offshore bus regions assigned to the onshore balancing authority/state/interconnection which they are connected to. This means during clustering the nodes for an offshore bus can be combined with the corresponding onshore bus. Alternatively we could keep these separate, but to do so I would need to add some dummy generators of >0 capacity to work with the clustering algorithm from EUR.

Issue #63, opened 2023-09-25
NREL ATB Fuel Costs for Nuclear and Biopower 


trevorb1:
### Feature Request

Right now, fuel costs for natural gas are pulled from the EIA. However, all other fuel costs use the standard values found in the `technology-data` download. This data is mostly based on European data. The [NREL ATB](https://atb.nrel.gov/) will provide fuel costs for Biopower and Uranium that we should pull in. 

### Suggested Solution

Much of the logic is already there to build the fuel costs `core_metric_parameter` key. It is mostly just mapping the values over in `add_electricity.py`

PR #126, opened 2023-11-10
New plot, battery fix, config clean up.

ktehranchi:
- [x] Adds brownfield map plot
- [x] constrains p_nom_min of batteries to p_nom
- [x] cleans up `config.default.yaml`
- [x] assigns haversine lengths to lines. (More work in-comming soon on TX capital costs)
- [x] Fixes monthly marginal cost data scaling issue

## Checklist


ktehranchi:
@trevorb1 

Constraining the p_nom_min for batteries to p_nom was the fix for batt dispatch. Issue now is that nuclear isn't being dispatched.... need to investigate... I think its the same marginal cost issue we had with CCGT and OCGT

ktehranchi:
Update: found an issue here- I changed the multiply to divide.... This explains the low cost.
We want: $/MCF * (1 MCF / 1.036 MMBTU) * (1 MMBTU / 0.293 MWh) = $/MWh

https://github.com/PyPSA/pypsa-usa/blob/134d2511bbdf1680002cd12a5c58c166142c93b5/workflow/scripts/add_electricity.py#L335-L338

trevorb1:
Nice catch!

trevorb1:
On the point of nuclear dispatch, I will also look at issue #63 today which finishes the mapping of costs from the ATB.

PR #129, opened 2023-11-23
Economic Retirement 

### LINKED ISSUE #66: Optimal System Capacities not calcualted
### Checklist

- [X] I am using the current [`master`](https://github.com/PyPSA/pypsa-usa/tree/master) branch
- [X] I am running on an up-to-date [`pypsa-usa` environment](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

### The Issue

As per [this comment](https://github.com/PyPSA/pypsa-usa/pull/44#issuecomment-1732433554), all `generators.p_nom_min` are set to the `generators.p_nom` value. This is to prevent the selling of assets at a negative cost. 

As elaborated on in [this comment](https://github.com/PyPSA/pypsa-usa/pull/44#issuecomment-1734120637), this means that we will not capture the optimal (or "true") amount of capacity in the system.

### Steps To Reproduce

1. Run the default workflow with `snakemake -j6`
2. Check the `n.generators.p_nom_opt` values and they will be at least as big as `n.generators.p_nom`

image

### Expected Behavior

I would expect to see that if the optimal system used (for example) zero capacity of `oil`, for the results to reflect that in `p_nom_opt`

### Error Message

_No response_

### Anything else?

@ktehranchi suggests a solution in [this comment](https://github.com/PyPSA/pypsa-usa/pull/44#issuecomment-1734120637).

I think before implementing this solutions (which I am not against!), it may be worth exploring how PyPSA-Eur deals with this issue. My concern is that adding in dummy technologies will add in a bunch of new constraints to an already big model! So if there is a way to do it without dummy techs, we may want to explore that!

Or, maybe we add in the dummy techs just to get it working, then try to optimize the implementation later down the line

trevorb1:
Closes #66 #130

In this PR I add the function `correct_extendable_generators()` to the `add_extra_components` module. This function adds logic for economic retirement of generators. Specifically, it: 
1. Creates duplicate generators for any that are tagged as extendable. For example, an extendable `CCGT` generator will be split into `CCGT` and `CCGT new` 
2. `capital_costs` of existing extendable generators are replaced with fixed costs 
3. `p_nom_max` of existing extendable generators are set to `p_nom`
4. `p_nom_min` of existing and new generators is set to `0`

## Checklist


trevorb1:
Getting this weird issue where it is cheaper to sell off some existing renewable generators and rebuild in places with better potential. 

image

This also happens at a bus level for solar, but at a system level it ends up building way more

trevorb1:
Oh. Just kidding. Im an idiot and forgot to set the capacity factors of the new solar and wind plants.

trevorb1:
@ktehranchi So I couldn't actually find where `pypsa-eur` implemented economic retirement, so I implemented it as we discussed yesterday to keep this moving forward. Two questions for thoughts tho: 
1. I have the option to select `economic` or `technical` retirement in the config file. You think this is okay, or should we simplify this and only give the option for one retirement method? I lean towards keeping as an option (as thats whats implemented), if we choose to force `economic` retirement, it may be good to revert the change of setting `p_nom_min` as `p_nom` earlier in the workflow for clarity 
2. Currently, all extendable generators (including wind and solar) have dummy generators implemented. I guess in reality, it probably wont make sense to retire wind and solar, so I am wondering if in an effort to reduce number of decision variables the economic retirement should only be applied to conventional generators?

trevorb1:
Still a draft as docs and plotting need to be updated to reflect the above questions/discussion

trevorb1:
Forgot to add in the marginal cost of new generators. Will need to add that still

ktehranchi:
> @ktehranchi So I couldn't actually find where `pypsa-eur` implemented economic retirement, so I implemented it as we discussed yesterday to keep this moving forward. Two questions for thoughts tho:
> 
> 1. I have the option to select `economic` or `technical` retirement in the config file. You think this is okay, or should we simplify this and only give the option for one retirement method? I lean towards keeping as an option (as thats whats implemented), if we choose to force `economic` retirement, it may be good to revert the change of setting `p_nom_min` as `p_nom` earlier in the workflow for clarity
> 2. Currently, all extendable generators (including wind and solar) have dummy generators implemented. I guess in reality, it probably wont make sense to retire wind and solar, so I am wondering if in an effort to reduce number of decision variables the economic retirement should only be applied to conventional generators?

1: Yes i like the idea of having this in the configuration file.
2: We already had dummy generators for wind and solar generators at each eligible node in the network (barring land use restrictions), so no need to add those again.

trevorb1:
Right. I guess more what I was thinking that we are preventing the model from selling off existing wind/solar generators, even though we tag this as economic retirement. But I agree, that probably wouldnt make too much sense in reality. Might just add a note in the docs to note that economic retirement does not apply to solar/wind. Thanks!

trevorb1:
Pretty sure this is ready to get merged, I will just do the following checks tomorrow: 
- [ ] Run in retirement mode `technical`
- [ ] Run in retirement mode `economic`
- [ ] Compare `technical` retirement mode results to main branch results, as they should be the same

trevorb1:
There is a real minor difference in production results. I wait to discuss during our next meeting before merging in.

ktehranchi:
I just noticed that since pypsa version 0.25.0 there is a new field for generators called `stand_by_costs`.... which we could be using instead to assign our FOM cost to! 

If we assigned fixed OM to the stand-by cost, and just capex for new capex build, we would still need to implement the split for new and old generators in order to track retirements separately. but we should update out pypsa version and add the standby costs for FOM split from CAPEX

trevorb1:
@ktehranchi, Im not sure this will totally work? Stand by costs are only tracked if the generator/link is tagged as commitable? Therefore, we wouldn't actually be adding in any FOM since we arnt running unit commitment?

ktehranchi:
Where do you see that its only read if its committable? I can't find that.

trevorb1:
@ktehranchi I think this is ready to get merged now. Can you please just do a quick run to make sure everything works okay, as per our conversation earlier this week? 

Also, I left out the `stand_by_costs` conversion for now. We can revisit this in a new issue ticket if needed.

ktehranchi:
Looks good, just changed the docs description a little and default config. hit the merge when you're ready.

PR #152, opened 2024-01-25
Electricity generation data validation script

nomineb:
Closes # (if applicable).

## Checklist

- [✓] I tested my contribution locally and it seems to work fine.
- [✓] Code and workflow changes are sufficiently documented.


degleris1:
Nice! @nomineb Can you revert the changes to

- `config.default.yaml` - let's leave this as is. 
- `solve_network.py` - probably best we hold off on the transport model for now. Do keep the script somewhere for later use.
- `data_validation.ipynb` - I think you can move this to `pypsa-usa/notebooks` and it should be ignored auotmatically?

PR #154, opened 2024-01-31
Extraction of EIA Data

trevorb1:

In this PR I implement logic to extract data from the [EIA API](https://www.eia.gov/opendata/browser/). My thoughts are we are starting to have a bunch of different data come from the EIA, and standardizing this would be useful. 

@ktehranchi your thoughts on this would be great!! :) 

### Example: 
- See module header for example
- The code shows how I envision the implementation (but am very open to other implementations!) 
- Extraction of fuel costs by fuel (gas and coal) and industry is implemented 
- Gas imports and exports by point of entry is implemented 

### Benefits: 
- Proposed code is extensible to add new API calls for specific data. The user really just needs to add the API call and raw data formatting logic. 
- Control of output data formats. Using the API, it is much easier to standarize dataframe structures compared to direct URL dowload (where formatting seems much more all over the place) 
- Separation of EIA data from the PyPSA-USA workflow. This allows us to easily test our EIA data processing code. 
- Calling of data will be easier - for example, getting time dependent fuel costs for different years will be cleaner, as all this logic is held in the totally independent `eia.py` module 

### Drawbacks: 
- Requires an API key. While this is free and very quick to get through the EIA (see [here](https://www.eia.gov/opendata/register.php)), it is an extra step. 
- Transferring existing EIA data to this will should be done. I see the following that can be extended to this:
    - Historical load data
    - Gas storage and production facilities 
    - Powerplant capacity 
- API calls are limited to 5000 lines. So we will need to sort out a way to extract hourly data - this is a very solvable problem using `offset` in the api call, just needs to be implemented. 

## Checklist

### DIFF
--- a/workflow/config/config.api.yaml
+++ b/workflow/config/config.api.yaml
@@ -1,0 +1,4 @@
+# holds user api keys
+
+apis:
+  eia: 
--- a/workflow/scripts/eia.py
+++ b/workflow/scripts/eia.py
@@ -1,0 +1,398 @@
+"""
+Extracts EIA data 
+
+Examples: 
+>>> costs = FuelCosts("gas", "power", 2020, "xxxxxxxxxxxxxxxx")
+>>> costs.get_data()
+
+                                           series-description  value  units   state  
+period                                                                               
+2020-01-15  U.S. Natural Gas Electric Power Price (Dollars...   2.74  $/MCF    U.S.  
+2020-02-15  U.S. Natural Gas Electric Power Price (Dollars...   2.50  $/MCF    U.S.  
+...                                                       ...    ...    ...     ...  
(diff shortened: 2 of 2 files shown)


trevorb1:
I do also totally understand that this is getting towards what PowerGenome does, however, this is just limited to EIA data and therefore lighter weight.

trevorb1:
It may be nice to move the year argument to the `get_data()` method call, but thats an implementation detail I guess.

ktehranchi:
I agree this will be useful for historical fuel flow/price data. I am actually going to move away from EIA data for the historical demand, and back to the GridEmissions (https://github.com/jdechalendar/gridemissions/pull/20#issuecomment-1894595311). I was planning on getting this done today or tomorrow so we can probably remove the need for historical demand in this. 

I spent some time with the API in the past to get this data, but found the 6 mo files more simple solution.

Issue #145, opened 2023-12-06
StorageUnit names are not unique 


trevorb1:
### Checklist

- [X] I am using the current [`master`](https://github.com/PyPSA/pypsa-usa/tree/master) branch
- [X] I am running on an up-to-date [`pypsa-usa` environment](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

### The Issue

When I run the workflow, I get the following pypsa error from the `add_extra_components` rule

```bash 
INFO:pypsa.io:Imported network elec_s_40.nc has buses, carriers, generators, lines, links, loads, storage_units
ERROR:pypsa.io:Error, new components for StorageUnit are not unique
INFO:pypsa.io:Exported network elec_s_40_ec.nc has generators, storage_units, lines, links, stores, buses, carriers, loads
```

### Steps To Reproduce

1. Run the workflow from the main branch 

### Expected Behavior

StorageUnits should have unique names. Unless I am missing something?

Issue #58, opened 2023-09-25
Update Storage Costs to be USA data


trevorb1:
### Feature Request

To apply costs, `TechnologyData` (ie. the PyPSA-Eur data) is used a as a backbone. Then USA data (ie. ATB, EIA, ect...) replace the European data where ever possible. [This comment](https://github.com/PyPSA/pypsa-usa/pull/44#issuecomment-1731732162) has a list of major differences between the datasets. 

One exception is the cost data for electrical batteries, which was very different between PyPSA-Eur and PyPSA-USA. Right now, PyPSA-Eur data has been applied for this; this needs to be updated and an explanation of the cost difference should be found. 

### Suggested Solution

It may be a simple incorrect mapping of technologies, otherwise, looking at the assumptions for each technology is probably needed 

### Additional Info

Here is where the code skips applying the USA data 

https://github.com/PyPSA/pypsa-usa/blob/6f37e32de0b19b349e6ba134e1bcfa0f48e5b93a/workflow/scripts/build_cost_data.py#L264-L291

ktehranchi:
#148 Changed the way build_cost_data integrated the NREL ATB:
> Updates build_cost_data: a) The "default" technology class from the NREL ATB is selected before any user-set parameters b) The user can now add any technology from the ATB by adding their choice of technology in the constants file.

PR #180, opened 2024-02-21
Refactor/Update build_cost_data, updates battery storage data

### LINKED ISSUE #149: Make battery representation consistent
### Feature Request

Right now, battery storage is modelled both as links+store+bus and as storage units, depending on if it is extendable or not. This is a little confusing as you need to track both `battery`, `battery charger` and `battery discharger` to understand results. 

### Suggested Solution

Switch all batteries to links+store+bus, or to storage units. I dont really have a preference on which way, just perfer a consistent representation :) 

### Additional Info

_No response_

ktehranchi:
Closes #149 #145 #58 

- Refactor build_cost_data: Now the user will only need to specify pypsa-usa name, ATB dsiplay_name, and Capital Recovery Period. The Technology type name needs to be specified if it does not match the naming scheme in ATB as described below:
```
pypsa-name:{
    "display_name": "Utility-Scale Battery Storage - 4Hr",
    "crp":  ATB Capital Recovery Period value,
    "technology": ATB Technology Name- In most cases technology name is embedded in the display name preceeding ' - ', for example "Biopower - Dedicated" has a technology name of "Biopower". In the cases of Coal and Natural gas this is not true, so the technology name must be defined. If adding new technologies- keep an eye on the logger.warnings and check the ATB directly.
    "scenario": ATB Technology Cost Scenario- "Moderate" used if not defined,
    "core_metric_case": ATB Core Metric Case- "Market" used if not defined,
    }
```
- #149 : Batteries are all now added as StorageUnits
- #58: Batteries are now pulled from ATB using display name in form Xhr_battery_storage where X \in 2-10
- #145 : Should be fixed.
- Fixes round trip efficiency of batteries, set to 0.9

## Checklist


workflow/scripts/add_extra_components.py
```diff
@@ -74,7 +73,8 @@ def attach_storageunits(n, costs, elec_opts):
     lookup_dispatch = {"H2": "fuel cell", "battery": "battery inverter"}
 
     for carrier in carriers:
-        roundtrip_correction = 0.5 if carrier == "battery" else 1
+        max_hours = int(carrier.split("hr_")[0])
+        roundtrip_correction = 0.5 if "battery" in carrier else 1
```
trevorb1, line 77: Maybe a silly question, but where is this coming from?

workflow/scripts/constants.py
```diff
@@ -463,280 +463,112 @@
 If you want to use the default ATB technology, the minimum must be defined: pypsa-name, technology, crp.
 If a default classificiation is not defined by ATB, then the remaining fields will be used to extract data from the ATB.
 pypsa-name:{
-    "technology":"ATB-Name",
-    "name":"ATB-tech-abreviation",
-    "alias":"ATB-tech-alias.",
-    "detail":"ATB-tech-detail",
-    "crp":"ATB-crp",
+    "display_name": "Utility-Scale Battery Storage - 4Hr",
```
trevorb1, line 466: for consistency with rest of workflow, should `display_name` be `nice_name`? Or is there a reason for the difference?

trevorb1 (approved):
@ktehranchi lgtm! I agree with your choice of removing the defaults from the atb options; much cleaner :) 

Only change I made was to convert a print statement to a logging statement

workflow/scripts/build_cost_data.py
```diff
@@ -49,7 +49,7 @@ def build_core_metric_key(
     ----
     Will not work with Debt Fraction
     """
-
+    print(f"building core metric key for {core_metric_parameter}, {technology}")
```
trevorb1, line 52: Change to logging?

Issue #207, opened 2024-02-28
Add GitHub Action for testing of Networks


trevorb1:
### Feature Request

Similar to PyPSA-Eur implementation [here](https://github.com/PyPSA/pypsa-eur/blob/master/.github/workflows/ci.yaml), adding in a GitHub action to test network configurations would be good.

ktehranchi:
9777d41d260933686baf58d11e56e377063f7e90 created main.yml file for github CI actions. Pulled directly from PyPSA-EUR with modification for our repo.

ktehranchi:
#208 fixed the issue with the action not running.... Now I'm looking at the way we call the .test.sh script.... currently when you want to run a specific config.****.yaml you need to change the snakemake file to uncomment out the config you want to use...

We need to change this so the config is called directly from the snakemake run command.

trevorb1:
While this does not test the networks, the Calliope team showcase a nice way of validating all configuration files against standardized schemas, [here](https://github.com/calliope-project/euro-calliope/blob/develop/.github/workflows/schemavalidation.yaml)

trevorb1:
Chatting with @ktehranchi, a potential way of testing workflow functionality is to:
1. For each test config, supply the existing intermediate network files
2. Generate new network files for the proposed changes (not sure how we go about this)
3. Test dataframes against one another 

```python 
import pypsa 
from pandas.testing import assert_frame_equal

n_fixture = pypsa.Network(fixture.nc)
n_new = pypsa.Network(new.nc)

assert_frame_equal(n_fixture.loads_t["p_set"], n_new.loads_t["p_set"]
```

shankarkarki:
Is this still opened? I think I can work on this.

trevorb1:
@shankarkarki - it is still open! It does feels like a few things have been discussed in this thread, that may be worth breaking out: 
- The testing for valid configuration files (see also issue #382)
- The testing of modules
- The testing of the entire workflow (ie. given specific config files, a workflow should run) 

May be worth breaking this out into separate issues to facilitate discussion on each topic. Replicating the PyPSA-Eur testing of the workflow (ie. item 3) for PyPSA-USA would probably be doable! But maybe @ktehranchi has other thoughts.

ktehranchi:
@shankarkarki let me know if you would like to chat about this issue! I agree with trevor on breaking out this ticket.  A good place to start would be to reference the testing set up in pypsa-eur for inspiration

[...]

PR #208, opened 2024-03-03
Updates GitHub Action

trevorb1:

Updates the testing workflow file. I believe no configurations met the imposed filtering options on what os to run on. I have enabled running on main branch Linux. 

Should address comment in issue #207. Not tagging it as a linked PR in case other items still need to be addressed with the action 

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
       max-parallel: 1
       matrix:
         os:
-        # - ubuntu-latest
+        - ubuntu-latest
         - macos-latest
         # - windows-latest
         inhouse:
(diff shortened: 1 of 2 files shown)


ktehranchi:
I'm disabling the action until one of us spends time to properly set it up.

trevorb1:
@ktehranchi Agreed! Maybe a silly question, but whats the goal of the action? 
- To check if the snakemake workflow is still functional? In which case we can just use the dry run flag to test all the rules still connect? 
- To check if the snakemake workflow will run top to bottom? In which case we are having to create a seperate config file for every option in our config file? If this is the goal, I wonder if it is better to just test each script separately with simple network fixtures? 
- To check if the workflow will make choices we expect to confirm overall functionality? Like if we make OCGT super cheap, we expect all investments to be in OCGT. I think this is great, but I guess this will require some sort of additional logic on the results unless we just do visual inspections?

trevorb1:
closing as this is outdated

PR #227, opened 2024-03-20
Updates Shapes, Validation Plots, Simplification Method

ktehranchi:
- Updates GIS Shapes to improve validation / historical dispatch results. No load shedding for 100 node WECC, and very very little at 300 Node :)
- Changes simplify network method to use more of the TAMU base network data. Previous method followed PyPSA-EUR method of simplification which mapped lines to line_types then mapped to the number of equivalent lines at 230 kV. Now I'm using per-unit conversion system which uses all of the parameters available in the TAMU/BE network, and we retain the original transfer capacities of the network.
- Cleans powerplant data for a few mis-classified plants. 

## Checklist


ktehranchi:
@casinocullen  This PR will likely fix the ercot load shedding issues. I can run a 100 node western without any shedding- and very little at 300.

Issue #235, opened 2024-03-21
In the function `format_eia_api_data`, many datasets are missing US-level data (e.g. ng_industrial_price) thus will cause an error in the lambda function


casinocullen:
### Checklist

- [X] I am using the current [`master`](https://github.com/PyPSA/pypsa-usa/tree/master) branch
- [X] I am running on an up-to-date [`pypsa-usa` environment](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

### The Issue

Potential solution is to add  if bool(usa_average) else df["value"] in the end of `apply`
```
    df["value"] = df.apply(
        lambda x: x["value"] if not x["use_average"] else usa_average[x["period"]], 
        axis=1,
    ) if bool(usa_average) else df["value"]
```

trevorb1:
Good call, @casinocullen! This is addressed with the `eia.py` module, and any missing data is replaced with USA average. This function should probably be removed. Relates to Issue #231!

PR #249, opened 2024-04-07
Adds Fuel Plot Graph

### LINKED ISSUE #230: Compare Fuel Costs
### Feature Request

Compare the raw fuel prices for natural gas to determine where the differences are coming from. For example, CAISO has a huge spike, while EIA does not for washington. Its hard to tell if this is a fuel price issue or an issue with generators and this is comparing marginal cost 

image

### Suggested Solution

_No response_

### Additional Info

_No response_

trevorb1:
**DRAFT PR STILL**

Closes #230 

In this PR I have added a graph to plot fuel costs. Since the marginal costs include a flat VOM cost, I have just taken 2030 VOM values to remove. So while this is technically an estimate of fuel costs, it will be a close approximation. 

## Checklist


trevorb1:
There is something weird going on with fuel costs I believe. 
1. OCGT and CCGT prices are not coming out to be the same. Since fuel costs will not have efficiencies or anything applied, they should come out to the same. 

image

2. The fuel costs in general are not lining up with what comes out of the EIA module. For example, in Arizona 

image

Gas prices should be 13.706 $/MWh

image

Coal prices should be ~0.70 $/MWh (idk, this one is really off - also see issue #250)

image

ktehranchi:
I was looking at similar issues this last week regarding mis-classification of plants between oil coal and NG, but didn't pick up on this issue.

1. It looks like the efficiencies are still included in the table you posted since CCGT is lower than OCGT? Maybe the function for pulling Fuel Costs is off? In the PR I am going to submit, I add a new fields for generators that captures the Avg Fuel Cost and the FOM cost- just so we can plot this and have the data readily accessible.
2. Coal Marginal Costs should be in the ~15 $/MWH range... (reference Azevedo Simple Dispatch paper). And that's what I am getting for final marginal costs on my end.

I'll try to get my PR in before we meet Wednesday

trevorb1:
hmm... probably an issue with my get fuel costs function, then. Let me review this. Thanks!

trevorb1:
Okay, Im pretty sure the `get_fuel_costs` function is working correctly, now. I accidentally grabbed the Eur fixed VOM costs, rather than the ATB ones; this has been corrected. Still doesnt address the issue that CCGT and OCGT are different (although closer now)

image

image

ktehranchi:
hmm and confirming you used these FOMS ?     
```
"CCGT": {  # natural gas
        "display_name": "NG Combined Cycle (F-Frame)",
        "technology": "NaturalGas_FE",
        "crp": 30,
    },
    "OCGT": {  # natural gas
        "display_name": "NG Combustion Turbine (F-Frame)",
        "technology": "NaturalGas_FE",
        "crp": 30,
    },"
```

Could the difference have to do with clustering? like maybe the capacity weighted averages for MC during clustering isn't applied for eff? Seems unlikely.

trevorb1:
> hmm and confirming you used these FOMS ?

Yeah, I pulled them from the `resources/costs_2030.csv` file which  should use that dictionary to extract the ATB data? The VOM has an ATB key value associated with it 

> Could the difference have to do with clustering? like maybe the capacity weighted averages for MC during clustering isn't applied for eff? Seems unlikely.

Maybe! Its also a little odd that new and old coal have different fuel costs. It is at the 3 decimal point, so maybe its just a numerical error, but then its weird how gas doesnt suffer from that.

trevorb1:
Oh, I wonder if this has to do with the efficiencies. 

If I look at combustion turbine in the ATB it shows a heat rate of `9.717 MMBtu/MWh`. This value is translated into the `costs_2030.csv` file okay. 

image

A heatrate of `9.717 MMBtu/MWh` should be an efficiency of `35%` (3412 / 9717 = 0.3511). But here it is showing the efficiency of `~40%` which is applied from when the generators are attached. I wonder if somewhere along the lines we end up using the efficiency in the `costs_2030.csv` file to do a calc instead of the individual generator (or clustered generator) efficiency? 

I took a quick look and didnt see anything, but may have missed it.

trevorb1:
Okay, I believe this may relate to the legacy code to get fuel costs, described [here](https://github.com/PyPSA/pypsa-usa/blob/eb6e96f9bb21de896e8d8c8a57c2471a44e25a12/workflow/scripts/add_electricity.py#L1390-L1406) and in issue #235. I will switch to fixing issue 235 and come back to this. 

https://github.com/PyPSA/pypsa-usa/blob/eb6e96f9bb21de896e8d8c8a57c2471a44e25a12/workflow/scripts/add_electricity.py#L1390-L1406

ktehranchi:
Our efficiencies are actually a mix of ADS, EIA historical for existing plants, and ATB for new build plants. So we would expect those to be different. You can rerun this with the new fields for avg heat_rate and fuel_costs

trevorb1:
Another issue, as pointed out by @ktehranchi, is that coal costs are not changing in the final network. Although, when they are applied in add electricity there is variance. 

image

trevorb1:
As per our (@ktehranchi and Franklyn) discussion today, I will go ahead and merge this. The minor differences in the gas prices between OCGT and CCGT are likely due to clustering. Moreover, while there is an artifact that coal prices are being read in correctly at the individual generator level, but by time the network is solved, the price is flat.

Both these issues probably deserve more exploration, but will move to a different issue ticket. 

The main additions from this PR are: 
- Addition of `get_fuel_price` function
- Plotting of fuel prices. A single figure is plotted where the first frame is system level price per fuel, then subsequent plots are individual fuels per bus.

Issue #232, opened 2024-03-20
EIA API key needed when modeling other than CAISO


casinocullen:
### Checklist

- [X] I am using the current [`master`](https://github.com/PyPSA/pypsa-usa/tree/master) branch
- [X] I am running on an up-to-date [`pypsa-usa` environment](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

### The Issue

When building NG prices, now the only options are getting data from EIA or CAISO. If no EIA API key is provided, no data would be prepared and cannot run the rest. 

Recommend to add a third option to provide a default NG price.

Also, seems like that data can't model period >= 2023, so if I am running period in 2023, there comes an error..

ktehranchi:
Do you think we should assert that `dynamic_fuel_price == true` the user must input an EIA API key? 

Or throw a warning and just fall back to the single average fuel cost

@trevorb1  thoughts?

casinocullen:
@ktehranchi That sounds reasonable to me

trevorb1:
@ktehranchi @casinocullen I agree! Asserting that the api key is provided is probably better than a warning, as it can be easy to miss with all the print outs :)

trevorb1:
Also, I add this documentation to the tasks in issue #220

PR #260, opened 2024-04-09
Updates dynamic pricing in `add_electricity`

### LINKED ISSUE #235: In the function `format_eia_api_data`, many datasets are missing US-level data (e.g. ng_industrial_price) thus will cause an error in the lambda function

trevorb1:
Closes #235 #232

In this PR I update how dynamic fuel pricing is extracted and applied. Specifically: 
- [x] Update EIA module to return pivoted data 
- [x] Replace `update_marginal_costs(...)` with `apply_dynamic_pricing(...)` which applies dynamic pricing to one geography (state, ba, reeds zone) at a time.
- [x] Update the call of `apply_dynamic_pricing(...)` to raise an error if no EIA API is given 
- [x] Update `build_fuel_prices.py` to return separate fuel files for different geographies. While this creates a little more clutter in the file tree, is greatly simplifies data processing imo. 
- [x] Remove any legacy code from `retrieve_cost_data.py` that has been replaced by the `eia` module. 

## Checklist


trevorb1:
A couple notes on the implementation:
- If `dynamic_fuel_price` is set to `false`, no fuel files are written out to the `resources/{interconnect}` folder
- If `dynamic_fuel_price` is set to `true`, and no API key is provided, an error is raised that is human readable (albiet, a little way into the workflow. but at least the error is clear) 
  - `AssertionError: Must provide EIA API key for dynamic fuel pricing`
  - A note on this, we can still bring in caiso data, but this will not allow it. We should move that to its own issue tho, as this PR is getting a little big. 
- If `dynamic_fuel_price` is set to `true`, and a API key is provided, fuel costs are written out at a per carrier and per geography level. Again, the motivation for this is while it results in a few more files, it makes the data processing easier.

trevorb1:
@ktehranchi, I tagged you as a reviewer, but I dont think a formal review is really required. Just since you did the original implementation of dynamic prices, wanted to give you a chance to give thoughts if you had any before merging. 

Functionality remains the same, but (I think) this implementation allows us to slot in new dynamic fuel pricing much easier

trevorb1:
Im gonna go ahead and merge this so I can continue with PR #249. If you have any concerns, please feel free to revert @ktehranchi! :)

PR #270, opened 2024-04-11
Adds configuration file templates 

### LINKED ISSUE #262: Ignore git tracking of all config files
### Feature Request

More of a question open for discussion, but it would be nice to add all configuration files to the git ignore. That way we have to be quite deliberate about updating them. I think this has three benefits: 
- Any keys will not accidentally be updated (ie. The EIA API key) 
- Simplifies switching between branches if you want to try config options on different features 
- Ensures our default config is tested and we know works 

Just a suggestion! If other people are not into this, thats fine too! :) 

### Suggested Solution

I thought I added the api config to the gitignore in PR #238, but mine is still being tracked. I guess I messed it up haha.  

### Additional Info

_No response_

trevorb1:
Closes #262

In this PR I have removed tracking of configuration files. Since (from what I saw) there is no option to have files in your repo that are not tracked when modified, I added template config files in the `repo_data/` folder. When first installing `pypsa-usa`, the user must run the new `init_pypa_usa.sh` script to copy the template files to the `config` folder. 

To summarize the changes: 
- config files have been moved to `repo_data/config`
- a script has been added to copy these files: `init_pypsa_usa.sh`
- all files in `workflow/config` are not tracked by git
- install instructions have been updated to tell the user to run the initialization script. 

Alternatively, we could probably use the [`onstart`](https://snakemake.readthedocs.io/en/v8.10.6/snakefiles/rules.html#onstart-onsuccess-and-onerror-handlers) argument from snakemake to check for the files beforehand. However, this may be difficult cause we dont really need all the config files for a run (for example, the `config.cluster.yaml` is not required for local runs).  

There are some git options (such as `assume-unchanged` or `skip-worktree`) that can be used to accomplish not tracking changes to git tracked files. However, this doesnt seem to be recommended practice. 

## Checklist


ktehranchi:
I think this is a great idea! I can modify my test config bash script so we can use it to double check if our changes match the template

PR #336, opened 2024-05-15
Index snapshots by planning horizon 

### LINKED ISSUE #335: Switch snapshots to use planning hroizon
### Feature Request

Change snapshot definitions so the datetime index uses the planning horizon year instead of the profile year for investment models. 

### Suggested Solution

_No response_

### Additional Info

_No response_

trevorb1:
Closes #335 

In this PR I have changed the indexing logic to be the planning horizon instead of the profile year. This includes: 
- Adding multi-index to all snapshot data (regardless if it is single horizon or not) in `build_base_network`
- Updating demand to export planning horizon data (note, not scaled data yet) - the `expand_demand` function is a little awkward, but it will allow for addition of scaling demand relatively easily. 
- Updating data in `add_electricity` to account for planning horizon (ie. renewables, fuel prices, ect...) through updated `broadcast_investment_horizon_index` function 

I have checked that both single horizon and multi-horizon data still works. Todo before merging: 
- [x] Confirm all plots still look correct 
- [x] Compare this PR against current develop branch to check nothing has changed (other than indexing) - but objective cost and other results should still be the same 

## Checklist

### DIFF
--- a/workflow/scripts/build_base_network.py
+++ b/workflow/scripts/build_base_network.py
@@ -823,10 +823,7 @@ def main(snakemake):
     planning_horizons = snakemake.params.planning_horizons
     sns = make_snapshots(sns_config, planning_horizons)
     n.snapshots = sns
-    if planning_horizons:
-        n.set_investment_periods(periods=planning_horizons)
-    else:
-        n.set_investment_periods(periods=n.snapshots[0].year)
+    n.set_investment_periods(periods=planning_horizons)
 
     # export network
     n.export_to_netcdf(snakemake.output.network)
--- a/workflow/scripts/build_base_network.py
+++ b/workflow/scripts/build_base_network.py
@@ -644,10 +644,14 @@ def make_snapshots(
         invest_periods = [pd.to_datetime(sns_config["start"]).year]
     sns = pd.DatetimeIndex([])
     for year in invest_periods:
+        start = pd.to_datetime(sns_config["start"])
+        end = pd.to_datetime(sns_config["end"])
+        year_diff = end.year - start.year
+        assert year_diff in (0, 1)
         period = pd.date_range(
             freq="h",
-            start=pd.to_datetime(sns_config["start"]).replace(year=year),
-            end=pd.to_datetime(sns_config["end"]).replace(year=year),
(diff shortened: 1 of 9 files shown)


trevorb1:
@ktehranchi, are you able to take a quick look (and maybe a quick run) to make sure this looks okay, please. Just want to make sure this stays true to your multi-horizon implementation implementation. 

Ive run a few different edge cases, and everything seems to work correctly! Comparing against the develop branch, the objective cost continually comes out ~4% different; however the results seem the same (just visually comparing graphs). 4% seems like a lot for simply numerical errors, tho. I can do a better exploration to see where the differences are coming from if you think this is needed.

workflow/scripts/build_base_network.py
```diff
@@ -796,15 +819,14 @@ def main(snakemake):
...
-        ),
-    )
+    sns_config = snakemake.params["snapshots"]
+    planning_horizons = snakemake.params.planning_horizons
+    sns = make_snapshots(sns_config, planning_horizons)
+    n.snapshots = sns
+    if planning_horizons:
+        n.set_investment_periods(periods=planning_horizons)
+    else:
+        n.set_investment_periods(periods=n.snapshots[0].year)
```
ktehranchi, line 829: When would panning_horizons ever be None here?

ktehranchi (approved):
Looks good! ran through and i think this makes more sense for indexing by planning horizon.

workflow/scripts/build_base_network.py
trevorb1: For validation runs, I believe! In that case, your planning horizon is empty and it just takes the snapshot year as the planning horizon. Justification for this was it is easier to manage always having multi-index, rather than separate logic for single vs. multi-index

ktehranchi: OOoo okay- for validation runs, I have been setting planning horizon to 2019 so either way works.

trevorb1: Fair point! have updated to require planning horizons. Porbably easier in the long run as well, if we alwasys know at least one planning horizon must be provided.

PR #358, opened 2024-06-07
Scales Demand 

### LINKED ISSUE #276: Apply scaling factor to loads
### Feature Request

PR #202 pulls in demand projections from the AEO and processes EFS data to show load growth rates per year. Logic needs to be added to apply these growth rates, as defined in the config file 

### Suggested Solution

_No response_

### Additional Info

_No response_

### LINKED ISSUE #339: Source EIA AEO from PUDL
### Feature Request

For future work that utilizes the AEO- we should go to the PUDL table that contains this data.

Data:
https://catalystcoop-pudl.readthedocs.io/en/nightly/data_dictionaries/pudl_db.html#core-eiaaeo-yearly-projected-generation-in-electric-sector-by-technology

Relevant modules:

- We'll need to access to this data in `build_demand` that's likely the place where accessing the sqlite db makes the most sense.

### Suggested Solution

_No response_

### Additional Info

_No response_

trevorb1:
Closes #339 
Closes #276 

In this PR I add sacling to demand. Specifically;
- If EIA data is used, data is scaled according to AEO data (now pulled from PUDL). 
- If EFS data is used, data is interpolated between the two closes EFS years
- A Demand Formatter class is added which handles scaling and formatting data, regardless of sector. 
- The demand formatter will instantiate a DemandScaler object which handles logic for scaling specific data

While this is not the most efficient implementation (as, for example, EFS data has to get read in once during the reading/writing, and again for scaling), I beleive this is clearer and more maintainable for our current workload. But happy to discuss this point! :) 

This PR also removes the option of selecting scaling method in the configuration file, as we should lock this to align data. 

## Checklist

### DIFF
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -88,7 +88,7 @@ wildcard_constraints:
 
 
 # Merge subworkflow configs and main config
-configfile: "config/config.validation.yaml"
+configfile: "config/config.default.yaml"
 configfile: "config/config.cluster.yaml"
 configfile: "config/config.common.yaml"
 configfile: "config/config.plotting.yaml"
--- a/workflow/repo_data/config/config.default.yaml
+++ b/workflow/repo_data/config/config.default.yaml
@@ -67,9 +67,7 @@ electricity:
     Link: [] 
 
   demand: 
-    profile: efs # efs, eia
-    scale: 1 # efs, aeo, or a number 
-    disaggregation: pop # pop
+    profile: eia # efs, eia
     scenario: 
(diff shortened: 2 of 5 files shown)


trevorb1:
Notes on this PR: 
- The first year must be an actual year of data to scale. For example, you cannot do a 2045 and 2050 EFS study, as 2045 demand data does not exist from EFS. You can do a 2040 and 2045 study, though. 
- EFS scaling can be improved to be by sector and region. Right now it is just linear interpolation for all data over the entire usa for the year

PR #420, opened 2024-09-24
Implements option to build individual/groups of States/Regions

### LINKED ISSUE #292: Add ability to build network for individual/groups of states
### Feature Request

For regional planning it would be useful to implement a feature that builds individual states rather than entire interconnections.

### Suggested Solution

- Filter states in `build_base_network`, `build_shapes`, and `build_bus_regions`
- Add electricity would need to only add plants that are located in the states

### Additional Info

_No response_

ktehranchi:
Closes #292:

This PR implements the option to model individual or groups of regions. I have moved model topology related configurations into the new section called `model_topology` which is used to specify the regions to include

### Example: Modeling California

To create a model that includes only California, you can specify the relevant ReEDS zone IDs (p8-11) as shown below. This will limit the spatial scope to the specified regions within California.

```yaml
model_topology:
   include:
      reeds_zone: ['p8', 'p9', 'p10', 'p11']
```

Alternatively, you can use the reeds_state: 'CA' option to achieve the same result by specifying the entire state. 

# Other changes:
- upgrades environment for pypsa to 0.30.1 - so update your env!
- addresses the many old deprecation warnings we had around the workflow
- Throws GIS related files into a subfolder for each resource group
- Includes WIP for county level aggregation

## Checklist


ktehranchi:
Hey @trevorb1 - let me know if this causes any conflicts with the county naming scheme in your sector work! idt it should. Also note the update to the env and the updated configuration files

docs/source/config-spatial.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Configuring Spatial Resolution
+
+PyPSA-USA gives you control over the spatial resolution of your transmission and resource networks, allowing for detailed or aggregated views depending on your modeling needs.
+
+### Transmission Network Selection
+You can specify the transmission network you want to use by setting the model_topology: transmission_network: option. There are two available options:
+
+- 'tamu': The synthetic BE-TAMU nodal network.
+- 'reeds': The ReEDS zonal network.
```
trevorb1, line 43: See note below about this

workflow/repo_data/config/config.default.yaml
```diff
@@ -18,6 +18,13 @@ scenario:
   sector: "" # G
   planning_horizons: [2030, 2040, 2050]
 
+# docs :
+model_topology:
+  transmission_network: 'tamu' # [reeds, tamu]
+  include: # mixed zone types not supported
+    reeds_zone: ['p8']
+    # reeds_state: ['CO']
```
trevorb1, line 26: Im not sure if I am missunderstanding, but I am having trouble getting these config options to work properly. If I change the transmission network to reeds, and set the topological_boundaries to either `reeds_zone` or `state`, these both fail. I run into the `n_clusters` assertion, and make sure I respect the clustering requirements.

workflow/scripts/build_renewable_profiles.py
```diff
@@ -226,7 +226,8 @@ def plot_data(data):
     cutout = atlite.Cutout(snakemake.input.cutout).sel(time=sns)
 
     regions = gpd.read_file(snakemake.input.regions)
-    assert not regions.empty, (
+
+    assert "x" in regions.columns and "y" in regions.columns, (
         f"List of regions in {snakemake.input.regions} is empty, please "
```
trevorb1, line 231: This caused the following error to appear when running the default configuration file. I had to remove the `offwind` and `offwind_floating` techs to get it to work. If this is required, we should probably just remove them from the default config, or (ideally?) correct this issue?

workflow/scripts/cluster_network.py
```diff
@@ -589,6 +598,13 @@ def plot_busmap_for_n_clusters(n, n_clusters, fn=None):
     else:
         n_clusters = int(snakemake.wildcards.clusters)
 
+    if transport_model:
+        logger.info(f"Using Transport Model.")
+        nodes_req = n.buses[f"{topological_boundaries}"].unique()
```
trevorb1, line 603: The topological boundary should be `state` or `reeds_zone` according to the config? But columns here are `country` and `county`? Not sure if I am missing something?

docs/source/config-configuration.md
```diff
@@ -215,7 +215,7 @@ Sector coupling studies are all under active development. More info to come!
 (clustering_cf)=
 ## `clustering`
 
-When clustering `aggregation_zones` defines the region boundaries which will be respected through the clustering process; State boarders, balancing authority regions, or REeDs shapes. This feature is important for imposing constraints (`opts`) which are defined over specific regions. For example, the data included in the model on interface transfer capacities are prepared for REeDs shapes but not states and BA regions. Moving forward we plan to use REeDs shapes as our default however we will maintain States and BA regions as well.
+When clustering `topological_boundaries` defines the region boundaries which will be respected through the clustering process; State boarders, balancing authority regions, or REeDs shapes. This feature is important for imposing constraints (`opts`) which are defined over specific regions. For example, the data included in the model on interface transfer capacities are prepared for REeDs shapes but not states and BA regions. Moving forward we plan to use REeDs shapes as our default however we will maintain States and BA regions as well.
```
trevorb1, line 218: is balancing authorities supported? Config file only shows state and reeds_zone in the comment

workflow/repo_data/config/config.default.yaml
trevorb1: Also the `ValueError` in `build_bus_regions` says something different 

```bash
ValueError: zonal_aggregation must be either balancing_area, country, reeds_id, or state
```

trevorb1 (changes requested):

docs/source/config-configuration.md
trevorb1: Also the ValueError in build_bus_regions says something different

```
ValueError: zonal_aggregation must be either balancing_area, country, reeds_id, or state
```

workflow/repo_data/config/config.default.yaml
ktehranchi: Topological boundaries shouldn't change from reeds_zone. I was considering getting rid of that config but whated to check with you. Okay, I'm able to run with  `topological_boundaries: reeds_zone ` so lets review this. The n_clusters assertion is required if you use `transmission_network: reeds`

trevorb1: You have to set `transmission_network` to `tamu` if you want to use the `include` functions? It wont work if its `reeds`?

docs/source/config-configuration.md
ktehranchi: removing this so we can move forward with only reeds_zone and county supported.

workflow/repo_data/config/config.default.yaml
ktehranchi: resolving since we realized we need an assertion that the inclusion zones exist within the interconnect

workflow/scripts/cluster_network.py
ktehranchi: Changing to only list reeds_zone as elegible topological boundary

workflow/scripts/build_renewable_profiles.py
ktehranchi: created issue #426

Issue #282, opened 2024-04-19
Package PyPSA-USA


trevorb1:
### Feature Request

I need to use the `INTERCONNECT_2_CODE` constant in the snakefile, as some demand data will come in at a state level. It would be nice to be able to access this constant directly, instead of having to duplicate it in the snakefile. It would be nice to package pypsa-usa to allow for this. For example, we could just call: 

```python 
from pypsa_usa.constants import INTERCONNECT_2_CODE
```

### Suggested Solution

Ive use [pyscaffold](https://pyscaffold.org/en/stable/) and [hatch](https://hatch.pypa.io/latest/) to do this before. They both allow for converting existing projects to packaged projects. Im not saying we push pypsa-usa to pypi, but allowing for local install via `pip install .` (or `pip install -e .` for developers) would be nice. Also, I am def not super experienced with packaging projects, so there may be a better tool to do this. imo hatch was a little easier to work with though, due to less setup files.

trevorb1:
Another advantage of packaging is it allows us to group our modules a little better. For example, we can group all `plot_*.py` modules into a `plotting/` directory

trevorb1:
When sorting through this, would be good to figure out how to update the automatic pre-commit PRs so both the master and develop branch update at the same time

PR #425, opened 2024-09-27
Adds developer install and basic pyproj

ktehranchi:
Added pre-commit hooks and a developer pip install so developer install is more clearly stated.


trevorb1:
@ktehranchi; Thanks for getting the package build started! 
1. [ I made a suggestion here](https://github.com/PyPSA/pypsa-usa/commit/02333dbfcb9971c0f8d8a40995d50dc26f0eaf60) on moving the developer installs into the contributing guide. Motivation was just to keep all developer stuff separate from users. Feel free to pull in the suggestions, or leave as is :) Im not fussed either way. 
2. I think if we start going down the route of including packaging files (which I am on board with), we should really look at using a tool such as hatchling (see issue #282 as well). [I had to add in the build directory](https://github.com/PyPSA/pypsa-usa/commit/bbc88f15a02db854675cf71b3d0ee9ad28e43ff1) to avoid install errors (see below). Moreover, if we use hatchling (or similar) it can help manage versioning. 

Im not against merging in what we have for packaging right now! I have just found in previous projects that using a tool to help manage packaging is very helpful! 

This is the error I get if I dont add the build directory: 

```bash 
Obtaining file:///home/trevor/master/pypsa-usa
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [14 lines of output]
      error: Multiple top-level packages discovered in a flat-layout: ['data', 'workflow'].
      
      To avoid accidental inclusion of unwanted files or directories,
      setuptools will not proceed with this build.
      
      If you are trying to create a single distribution with multiple packages
      on purpose, you should not rely on automatic discovery.
      Instead, consider the following options:
      
      1. set up custom discovery (`find` directive with `include` or `exclude`)
      2. use a `src-layout`
      3. explicitly set `py_modules` or `packages` with a list of names
      
      To find more information, look for "package discovery" on setuptools docs.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build editable did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

ktehranchi:
Per discussion:
1- I have merged trevors changes on the other branch `upstream/kt/pyproj`
2- @trevorb1 will address hatchling in another branch/PR

PR #443, opened 2024-10-17
Adds Heat Pump Constraints 

### LINKED ISSUE #430: Heat Pump modelling for heating/cooling
### Feature Request

Right now, HPs only contribute to meeting heat demand, but this needs to be corrected so they can also contribute to cooling demand. 

### Suggested Solution

The cleanest solution would be to introduce a binary variable so HPs can only contribute to one load (heat/cool) at a time; but I def dont want to do this since the sector side is already getting quite heavy. 

Instead I propose the following: 
- If HPs are allowed, two HP links are added from the electrical bus. One link will go to the heat sector, one will go to the cooling sector
- The capacity of these will be constrained to be the same within solve network 
- The generation on the combined links at each timeslice will be constrained to be less than or equal to the max timeslice generation of only one of the heat pumps

This will allow HPs to be used to meet both heating and cooling demand within a single timeslice, but since we are aggregating (and not modelling individual units) this should be fine. 

### Additional Info

_No response_

### LINKED ISSUE #435: Constrain GSHP Capacity
### Feature Request

PR #434 introduces the ability to combine rural and urban areas. This will result in over investment of GSHP, which should not happen in urban areas. We can pull in population braeakdowns still to constrain the amount of generation allowed to meet service sector demand from GSHPs

### Suggested Solution

_No response_

### Additional Info

_No response_

trevorb1:
Closes #430 
Closes #435 

In this PR I have:
- Added functionality for HPs to meet both heating and cooling loads
  - The HP to meet heating demand is unchanged 
  - A dummy HP is added to go between electricity and cooling buses. This HP has the same COP, but has zero capex
  - The two HP links are constrained to have the same capacity 
  - The summed generation of the two HPs at each timeslice must be less than or equal to the max generation of a single HP (ie. Heating HP capacity times the COP) 
- Constrained max GSHP capacity for studies that group urban/rural into a single node. 
  - The ratio of urban to rural population is read in
  - The constraint then enforces this ratio between ASHP and GSHPs 

## Checklist


trevorb1:
Before merging, I need to do a little more testing with different network configs.

PR #525, opened 2025-02-07
Adds Power Sector Demand Response 

### LINKED ISSUE #523: Power Sector Demand Response
### Feature Request

Add power sector level demand response. I propose (at least to start) to follow the price based implementation in the sector system, allowing us to use the same constraints. 

### Suggested Solution

See sector implementation here: 

https://github.com/PyPSA/pypsa-usa/blob/568d664705443fcbe1dff0e86aa94741293353fd/workflow/scripts/build_electricity_sector.py#L84-L221

### Additional Info

_No response_

trevorb1:
Closes #523 

In this PR I add demand response at the power sector level. This can be applied to both electricity only studies, but also sector studies. The implementation follows the same that is used in sector studies, where a cost is incurred to store one unit of energy per each hour. Details on the implementation can be found [here](https://pypsa-usa.readthedocs.io/en/latest/data-sectors.html#demand-response). 

For the capacity constraints, electrical only and sector studies are treated different: 
- Electrical only constrains against the load at the bus that DR is implemented at. 
- Sector studies constrain against the sum of flows leaving the bus that DR is implemented at. 

A couple notes:
- By default DR is turned off in the configuration file. 
- The DR components are added in the `add_extra_components` script. However, @ktehranchi if there is a better place to add this, please advise. 
- No plotting scripts have been added to electrical only studies for DR

## Checklist


trevorb1:
@ktehranchi, would you mind giving this a quick look over please; since this functionality can also be used for power sector studies. I did a couple of tests with electrical only studies, and it seemed to be working okay, tho.

ktehranchi:
looks good to me, my only question was have you tested this in multi-horizon models- re definition of the capacity constraint

trevorb1:
Confirmed this works correctly with multi-period studies, and does not change solution when dr is not implemented (ie. components and constraints not active).

PR #534, opened 2025-02-11
Ruff Linting & File Removal

ktehranchi:
Along with #533 #532: This PR reformats our code to include the new changes from ruff lintiing. 

@trevorb1 there were a few flags that I didn't want to run on sector related files which I have blocked at the top of those files with:
```
# ruff: noqa: RUF012, D101, D102, N818, D400
```

Everyone should update their current working branches with develop and run:
```
pre-commit clean && pre-commit install
```


trevorb1 (commented):
@ktehranchi; thanks for doing this, and putting those skips at the top of the sector files. I will lint them separately after this is merged. I haven't done a detailed review of everything, as Im assuming most of the changes are just formatting changes. 

Two quick comments;
- I believe in the `.pre-commit-config.yaml` we need to add in the ruff linting? I dont see that change here? Something like: 

```yaml 
repos:
- repo: https://github.com/astral-sh/ruff-pre-commit
  # Ruff version.
  rev: v0.9.2
  hooks:
    # Run the linter.
    - id: ruff
      types_or: [ python, pyi ]
      args: [ --fix ]
    # Run the formatter.
    - id: ruff-format
      types_or: [ python, pyi ]
```

- If we are committing to Ruff, may be worth adding this to the CI workflow? Something like in `pypsa-usa/.github/workflows/ruff.yaml`: 

```yaml 
name: ruff
on: [ push, pull_request ]
jobs:
  ruff:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: astral-sh/ruff-action@v3
        with:
          src: "./src"
```

Im also fine if we want to do these in a separate PR though, after sector has been linted.

ktehranchi:
@trevorb1 Cool, I actually already changed the pre-commit-yaml in a separate PR (to both develop and main) which I already merged, thats why you don't see it here.

I like the RUFF badge, I will add that haha

PR #581, opened 2025-03-25
Restores conditional for foresight option in sovle network 

trevorb1:
Closes #580 

## Changes proposed in this Pull Request

Restores the conditional in solve network that was removed in PR #576 

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Changed dependencies are added to `envs/environment.yaml`.
- [ ] Changes in configuration options are added in all of `config.default.yaml`.
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv`.

PR #582, opened 2025-03-27
Conditional for Sector multi_period_investment

trevorb1:
Hotfix for PR #581 

As per PR #581, this causes sector to crash, but without it...
> pypsa doesn't properly account for the build_year and lifetime of assets. When you leave the flag off, it ignores those fields, which was leading to single horizon problems being improperly built.

This PR adds it in for electrical, but not for sectors. This is a temporary fix. 

## Checklist


trevorb1:
@ktehranchi updated to not impact electricity studies

Issue #518, opened 2025-01-29
Myopic Implementation


lfreese:
### Feature Request

Develop a myopic implementation of foresight rather than perfect.

### Suggested Solution

Solution is in progress, modifying **planning_horizons**

lfreese:
Two continuing issues in myopic approach:

1) I've updated the n.add() for each component to be less verbose as `n.add(nm, df_idx, **df.loc[df_idx])` [line 1558], but this leads to Generators missing passed attirbutes from the df such as land_region, heat_rate, etc. because of the warning "WARNING:pypsa.components:Generator has no attribute land_region, ignoring this passed value."
The only solution I've found to this is to re-add in a `n.madd()` for only the Generators, with each individual passed value being named. `n.madd()` is deprecated in future pypsa versions so this is not ideal to have

2) Line 1594 `n.storage_units.loc[:,'state_of_charge_initial'] = n.storage_units_t.state_of_charge.loc[planning_horizon].sum()
` leads to an infeasible solve. The purpose of this was to carry over the state of charge from the previous time period and set that as the state_of_charge_initial for the next time period. 

Relevant code in https://github.com/lfreese/pypsa-usa/tree/issue-518

lfreese:
Problem 2 above was solved and has been pushed. Needed to be: `n.storage_units.loc[:,'state_of_charge_initial'] = n.storage_units_t.state_of_charge.loc[planning_horizon].iloc[-1]` to get the final snapshot state of charge

lfreese:
@ktehranchi have you pushed your changes to the opts (the SAFE/SAFER) ones you discussed? If so, I'll pull those and add the sns_horizon to them so that they work with myopic. If those are still being developed, I think this might be ready for a pull request. It doesn't do quite as well in the full year snapshot recreating the perfect foresight as it did in a 2-3 day snapshot, so I'm not sure if we feel like that should preclude pulling this, or if that's reasonable. 

Figures here show examples of the difference in production and capacity additions between a 2030 myopic and 2030 perfect foresight. Config file attached. 

[config.myopic.txt](https://github.com/user-attachments/files/19458365/config.myopic.txt)

![Image](https://github.com/user-attachments/assets/58582f69-2753-49be-97e8-1fc84302b261)
![Image](https://github.com/user-attachments/assets/1e0e574e-5f26-4b95-8d68-d48cd5974910)

The key gaps in production are coming in with coal, OCGT, and CCGT, for example:

[...]

Issue #150, opened 2024-01-18
Restructure Directory Tree


trevorb1:
### Feature Request

We dont really need the `workflow/` directory, and move up all the subdirectories. Moreover, we can probably get rid of the `report/` directory

PR #592, opened 2025-04-14
solve_network refactor, ERM, & opts testing 

ktehranchi:
- Refactors solve_network module to break opts into sub-modules which we import. Motivated by the unruly size the solve_network module had become. 
- Implements Energy Reserve Margin (ERM) opt. Follows implementation described in [Stephen & Kirschen 2024](https://ieeexplore.ieee.org/document/10667201)
- Implements test for ERM. I have added placeholders for other opts. The structure of the tests will reuse basic 3-bus test networks

## Checklist


ktehranchi:
@trevorb1 I'm down to merge this in after the v0.6.0 release. I'm just requesting review to see if you have any feedback on how I have restructured the opts and solve_network. No need to do detailed review of the new ERM.

workflow/scripts/test/conftest.py
```diff
@@ -0,0 +1,551 @@
...
+        "Carrier",
+        "battery",
+        co2_emissions=0,
+    )
+    n.add(
+        "Carrier",
+        "AC",
+        co2_emissions=0,
+    )
+    return n
```
trevorb1, line 303: I really like the `base_network()` fixture, but wonder too much is included in it? Carriers that are here, but only used in some constraints (ie. `nuclear`), component interactions that shouldnt happen are included (ie. lines and links both being used for power transfer), attributes that are only used for particular constraints are included globally (ie. `land_area`), ect. 

This can be an issue as then we are not isolating each test. For example, (a fictional case could be) if a constraint uses a shared function to check that component names are defined correctly and there is an issue with how a line is defined. All tests would fail cause lines are always included in the base network. However, if running a transport model, these tests should pass as lines are replaced with links. This can make diagnosing the problem harder. (I do concede that is is a fairly extreme edge case 😅)

Would it be better to have like a high level `base_network()` that has only components that are shared across all network configurations (ie. snapshots, buses, loads)? Then we can have a few "sub" base networks. For example, one that is transport or not transport; one that is electrical only vs sector (where, for example some generators are replaced with links); ect? 

Or maybe the base network is literally a 1, 2, or 3 node network that only defines the snapshots and buses. Then each constraint adds in exactly what it needs? ie. for some constraints, we dont even need multi-node networks, or storage units? 

Or is this creating too much structure?

workflow/scripts/test/conftest.py
```diff
@@ -0,0 +1,551 @@
...
+            )
+
+    snakemake = MockSnakemake()
+
+    yield config, snakemake
+
+    # Clean up temporary files
+    os.unlink(portfolio_standards_file.name)
+    os.unlink(rps_reeds_file.name)
+    os.unlink(ces_reeds_file.name)
```
trevorb1, line 464: Would it be easier to just have simple CSVs in a fixtures folder, as that how its actually used in the workflow? Not that there is anything wrong with your implementation! Would just get around having to do the `NamedTemporaryFile` I think?

workflow/scripts/test/test_content.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+
+    n.add(
+        "Load",
+        "load_east",
+        bus="bus3",
+        p_set=pd.Series(250, index=n.snapshots),
+    )
+
+    return n
```
trevorb1, line 244: Why cant you use the base network here, and build onto it like in the other tests?

workflow/scripts/test/test_content.py
```diff
@@ -0,0 +1,372 @@
...
+            )
+
+    snakemake = MockSnakemake()
+
+    yield config, snakemake
+
+    # Clean up temporary files
+    os.unlink(portfolio_standards_file.name)
+    os.unlink(rps_reeds_file.name)
+    os.unlink(ces_reeds_file.name)
```
trevorb1, line 322: Wasnt something similar already defined in conftest?

workflow/scripts/test/README.md
```diff
@@ -0,0 +1,129 @@
...
+  - Data storage
+  - Storage contribution to ERM
+  - Transmission line contribution to ERM
+
+- Tests for ERM constraints across multiple regions
+  - Constraint application in multiple regions
+  - Interconnection contribution to ERM
+
+- Tests for basic functionality of PRM constraints
+  - Constraint creation
```
trevorb1, line 29: Do we actually test the constraint creation? I may have missed it, but seems like we test constraint behaviour? I would consider testing constraint creation to be how linopy describes [here](https://linopy.readthedocs.io/en/latest/testing-framework.html#Testing-with-Linopy). (Sorry, this one is very nitpicky!)

trevorb1 (commented):
Hi @ktehranchi; this looks really really great! Thanks for staring this (long overdue) testing setup. Please see below for my comments. Also, a couple general questions: 

1. Im not sure why we need the `sys.path.append(os.path.join(os.path.dirname(__file__), ".."))` at the top of all the testing files? 
2. Are we wanting to add the tests to the GitHub Workflow as well? Or you think since each test does an optimization (ie, its kinda heavy for a test) we leave it up to the user to test locally? 
3. We should add that users should run tests in the contributing guide in the docs as well? (ie. the code snippet you have in the README.md copy/paste to the contributing guide)

workflow/scripts/test/README.md
ktehranchi: I used to test the constraint creation, so I should get rid of this from the README too.

workflow/scripts/test/conftest.py
ktehranchi: thats true, I will do that with some test data in the same format

workflow/scripts/test/conftest.py
ktehranchi: I like this idea, to have a high level base_network then some sub networks that are made to test different possible configurations.... I'll make this change, thanks!

workflow/scripts/solve_network.py
```diff
@@ -25,133 +25,47 @@
...
+from opts.sector import (
+    add_cooling_heat_pump_constraints,
+    add_demand_response_constraint,
+    add_ev_generation_constraint,
+    add_gshp_capacity_constraint,
+    add_ng_import_export_limits,
+    add_sector_co2_constraints,
+    add_sector_demand_response_constraints,
+    add_water_heater_constraints,
+)
```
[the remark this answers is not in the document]
ktehranchi: I don't think we do since these are just relative imports??

workflow/scripts/test/README.md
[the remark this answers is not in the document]
ktehranchi: oh yeah I guess normally the test would go in the base directory. I am going to leave restructuring the workflow for packaging for another PR. We had mentioned this previously in  #150

workflow/scripts/test/run_tests.py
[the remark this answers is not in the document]
ktehranchi: removed!

workflow/scripts/test/README.md
```diff
@@ -0,0 +1,129 @@
...
+## Requirements
+
+- **pytest**: The tests use pytest as the testing framework
+- **GLPK**: The tests use the GLPK solver by default
+
+Install the required packages:
+
+```bash
+pip install pytest
+```
```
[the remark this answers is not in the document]
ktehranchi: Removed this from readme, and added pytest to dev.yaml. It was already in `uv`

workflow/scripts/opts/sector.py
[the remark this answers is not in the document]
ktehranchi: yep direct copy and paste

workflow/scripts/solve_network.py
```diff
@@ -1742,6 +230,76 @@ def run_optimize(n, rolling_horizon, skip_iterations, cf_solving, **kwargs):
...
+                n.add(nm, df_idx, **df.loc[df_idx])
+        logger.info(n.consistency_check())
+
+        # copy time-dependent
+        selection = n.component_attrs[nm].type.str.contains("series")
+        for tattr in n.component_attrs[nm].index[selection]:
+            n.import_series_from_dataframe(time_df[tattr], nm, tattr)
+
+    # roll over the last snapshot of time varying storage state of charge to be the state_of_charge_initial for the next time period
+    n.storage_units.loc[:, "state_of_charge_initial"] = n.storage_units_t.state_of_charge.loc[planning_horizon].iloc[-1]
```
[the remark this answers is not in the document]
ktehranchi: This is just refactored/functionalized from the recent #518 !

ktehranchi:
> Hi @ktehranchi; this looks really really great! Thanks for staring this (long overdue) testing setup. Please see below for my comments. Also, a couple general questions:
> 
> 1. Im not sure why we need the `sys.path.append(os.path.join(os.path.dirname(__file__), ".."))` at the top of all the testing files?

This has to do with your comment on the packaging and structure of the repository. im going to leave this to address in #150 

> 2. Are we wanting to add the tests to the GitHub Workflow as well? Or you think since each test does an optimization (ie, it's kinda heavy for a test) we leave it up to the user to test locally?

I don't think the optimization is too heavy! the test networks are all <48 hours of data, and glpk can solve it extremely quickly. We can make it even lighter weight. I will try to set up the online testing at some point per #207 

> 3. We should add that users should run tests in the contributing guide in the docs as well? (ie. the code snippet you have in the README.md copy/paste to the contributing guide)

I will add this

ktehranchi:
Also thank you for the super detailed code review!!! @trevorb1

workflow/scripts/test/conftest.py
ktehranchi: created in #598

PR #601, opened 2025-05-16
Add north dakota node and fix pudl file path request

asiazzzhang:
Closes # (if applicable).

Added a bus in the empty county in North Dakota. Fixed the PUDL file path for EIA data.

## Checklist


ktehranchi:
@asiazzzhang  can you remove the godeeep.ipynb from this PR?

workflow/scripts/build_demand.py
```diff
@@ -1906,10 +1906,7 @@ def assign_scaler(self):  # type DemandScaler
             assert self.api, "Must provide eia api key"
             return AeoEnergyScaler(self.api)
         elif self.scaling_method == "aeo_electricity":
-            assert self.filepath.startswith(
-                "s3://pudl.catalyst.coop/",
-            ), "Must provide pudl S3 URL (s3://pudl.catalyst.coop/...)"
-            return AeoElectricityScaler(self.filepath)
+            return AeoElectricityScaler(snakemake.params.pudl_path)
```
trevorb1, line 1909: This now uses `snakemake` from the global namespace in the local method; is this change required? Just makes debugging a little more difficult!

trevorb1: Hmm... isnt `snakemake.params.pudl_path` just a string? So cant we still just pass in the pudl path as `filepath`? Like remove the assert, and leave `return AeoElectricityScaler(self.filepath)` and no need for the conditional?

trevorb1: We would just need to update the `demand_scale_file` function. 

https://github.com/PyPSA/pypsa-usa/blob/9b3e5237b6249881ef59f96e54339c5406a1a11a/workflow/rules/build_electricity.smk#L291-L310

trevorb1: It will get kinda messy since we are still fractured on what eia source we use (ie. PUDL does not have all projections for sector studies). but something like;

```
def demand_scaling_data(wildcards):
    ...
    elif profile == "eia": 
        if end_use == "power": 
            return config["pudl_path] # use versioned pudl data
        else:
            return [] # use data from EIA API directly 
        ...

Comment (ktehranchi) [reply workflow/scripts/build_demand.py:1909]: How about this to remove the snakemake from the local method:

```
    if (scaling_method == "aeo_electricity") and (end_use == "power"):
        demand_scale_file = snakemake.params.pudl_path
    else:
        demand_scale_file = snakemake.input.get("demand_scaling_file", None)

    demand_formatter = DemandFormatter(
        n=n,
        scaling_method=scaling_method,
        filepath=demand_scale_file,
        api=eia_api,
    )
```

Do we need to check for `and (end_use == "power")`?

PR #605, opened 2025-05-21
Fix EIA demand scaler bug, Missing SD zone

ktehranchi:
Replacement PR for #601


trevorb1:
lgtm! Thanks for the fix! :) See [this comment](https://github.com/PyPSA/pypsa-usa/pull/601#discussion_r2100860132) for my thoughts on your fix (which I agree with!).

PR #615, opened 2025-06-01
Fix Policy Opts for Myopic, Implements Trim Network feature

ktehranchi:
- Fixes issues with RPS, TCT, REM opts when using myopic algorithm (#518) @lfreese 
- Implements a new `trim` configuration for `model_topology`: Similar to `include` (which at some point I will deprecate), Trim splits a section of a network from the core network but retains all buses one-step outside of the network. Buses at those networks are equipped with imports generators set to the marginal cost of OCGT. Emission for the imports are set according to California's imports policy since I've developed this specifically for CA.
usage example to model CA at county resolution with reeds network while modeling external WECC regions as imports :
```
model_topology:
  transmission_network: 'reeds' # [reeds, tamu]
  topological_boundaries: 'county' # [county, reeds_zone]
  interface_transmission_limits: false
  include: # mixed zone types not supported
  aggregate: # eligible keys: [reeds_zone, trans_reg]
  trim: 
    zone: ['CA']
```
- Enables `nhours` param to be used for multi-horizon networks

### DIFF
--- a/workflow/scripts/add_extra_components.py
+++ b/workflow/scripts/add_extra_components.py
@@ -827,7 +827,7 @@ def trim_network(n, trim_topology):
     # Get OCGT generators and calculate average marginal cost
     ocgt_gens = n.generators[n.generators.carrier == "OCGT"]
     avg_marginal_cost = get_as_dense(n, "Generator", "marginal_cost").loc[:, ocgt_gens.index].mean().mean()
-    n.add("Carrier", "imports", co2_emissions=0.427, nice_name="imports")
+    n.add("Carrier", "imports", co2_emissions=0.428, nice_name="imports")
 
     # remove existing oneport components at bus
     for c in n.one_port_components:
--- a/workflow/scripts/opts/_helpers.py
+++ b/workflow/scripts/opts/_helpers.py
@@ -1,5 +1,6 @@
 import logging
 
+import linopy
 import numpy as np
 import pandas as pd
 import pypsa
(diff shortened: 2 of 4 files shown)


ktehranchi:
@trevorb1 I implemented this for my current project but I think it could be useful for your research issue regarding treatment of imports from modeling each BA/state seperately

trevorb1:
@ktehranchi thanks for this! I am doing something similar, but at ISO level (ie. your implementation is more general). A couple notes on the differences that you may have thoughts on? 
1. I pull in data on historical net-flows, and balance over one of day/week/month/year. The data I use is "core_eia930__hourly_interchange" from PUDL [here](https://viewer.catalyst.coop/). I guess this may difficult to generalize though, since this data does not come in at BA level. Moreover, this is only net, so Im not sure how sensitive it is to the choice of balancing period. 
2. I retain the capacity limits at a BA level into neighbouring ISOs. This is same as what you do, I believe! (But my implementation isn't integrated into PyPSA-USA yet, so yours here is better!). 
3. I pull in historical [EIA wholesale electricity market data](https://www.eia.gov/electricity/wholesale/) as the import/export price. 
4. I guess the key difference between our implementations is that I specifically don't want to account emissions associated with imports against the state emission budget, but do retain emissions associated with electricity generated for exports.

trevorb1:
I guess [EIA actually has state level electricity imports](https://www.eia.gov/opendata/browser/electricity/state-electricity-profiles/source-disposition?frequency=annual&data=net-interstate-trade;&sortColumn=period;&sortDirection=desc;), which could address my point 1 of the regions provided by 930 being too coarse.

PR #622, opened 2025-06-18
Add CCTS functionalities

ricnogfer:
Closes # (if applicable).

## Checklist

### DIFF
--- a/workflow/scripts/add_sectors.py
+++ b/workflow/scripts/add_sectors.py
@@ -771,10 +771,11 @@ if __name__ == "__main__":
 
     # add node level DAC capabilities
     if snakemake.config["dac"]["enable"]:
+        raise ValueError("DAC is not supported for Sector Studies. See https://github.com/PyPSA/pypsa-usa/issues/652")
         if snakemake.config["co2"]["storage"]:
-            logger.info("Adding node level DAC capabilities")
+            logger.info("Adding DAC capabilities")
             add_dac(n, snakemake.config, True)
         else:
-            logger.warning("Not adding node level DAC capabilities given that CO2 (underground) storage is not enabled")
+            logger.warning("Not adding DAC capabilities given that CO2 (underground) storage is not enabled")
 
     n.export_to_netcdf(snakemake.output.network)


ktehranchi:
Hey Ricardo- Taking a look at the changes, but could you please first re-base your changes on the develop branch? The diff is massive since I believe you were working off the release candidate.

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+    n.madd(
+        "Carrier",
+        ["co2"],
+        color=config["plotting"]["tech_colors"]["co2"],
+        nice_name=config["plotting"]["nice_names"]["co2"],
+    )
+
+    # add carrier to represent CC only (i.e. without S)
+    carriers = n.carriers.query("Carrier.str.endswith('CCS')")
+    if carriers.empty is False:
```
trevorb1, line 921: Kinda picky, but probably better to use `if not carriers.empty:` following the [Programming Recommendations in PEP8](https://peps.python.org/pep-0008/#programming-recommendations)? 

```suggestion
    if carriers.empty is False:
    if not carriers.empty:
```

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+            carriers.index.str.replace("CCS", "CC", regex=True),
+            color=carriers["color"],
+            nice_name=carriers["nice_name"].str.replace("Ccs", "Cc", regex=True),
+        )
+
+    # get CO2 intensity for gas and coal
+    gas_co2_intensity = costs.loc["gas"]["co2_emissions"]
+    coal_co2_intensity = costs.loc["coal"]["co2_emissions"]
+
+    if sector is True:
```
trevorb1, line 933: Similar PEP8 guideline. You can just write `if sector:` and drop the `is True` part

```suggestion
    if sector:
```

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+        links = n.links.index.str.contains("CCS")
+        if True in links:  # found links equipped with CCS
```
trevorb1, line 935:
```suggestion
        if links:  # found links equipped with CCS
```

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+            # specify links' bus4 to point to their respective CO2 capture buses
+            n.links.loc[links, "bus4"] = co2_storage.index + " co2 capture"
+
+            # calculate efficiencies
+            efficiency2 = []
+            efficiency4 = []
```
trevorb1, line 941: This is my fault for not documenting the different link destinations. But can we just add in comments here on the destination? (If my comment below is correct?) 

```suggestion
            efficiency2 = [] # to atmosphere
            efficiency4 = [] # to underground storage
```

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+    n.madd(
+        "Store",
+        co2_storage.index,
+        suffix=" co2 storage",
+        bus=co2_storage.index + " co2 capture",
+        e_nom_extendable=True,
+        e_nom_max=co2_storage["potential [MtCO2]"] * 1e6,  # in tCO2
+        marginal_cost=co2_storage["cost [USD/tCO2]"],
+        carrier="co2",
+    )
```
trevorb1, line 909: I would align the carrier suffix (ie. both `co2 storage` or `co2 capture`). But @ktehranchi may have different thoughts!

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+        marginal_cost=config["co2"]["network"]["marginal_cost"],
+        carrier="co2",
+        lifetime=config["co2"]["network"]["lifetime"],
+    )
+
+
+def add_dac(n: pypsa.Network, config: dict, sector: bool):
+    """Adds node level DAC capabilities."""
+    # generate node level buses to represent emitted, captured and accounted CO2 and links to represent DAC in function of whether network is based on sectors or not
+    if sector is True:
```
trevorb1, line 1121:
```suggestion
    if sector:
```

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+        # get DAC granularity/scope
+        granularity = config["dac"]["granularity"]
+        if granularity == "nation":
+            granularity = "node"
+            logger.warning(
+                "Nation level DAC capabilities is not applicable for a network based on sectors - defaulting to node level instead",
+            )
```
trevorb1, line 1128: Wait, why cant DAC be applied to remove emissions at a national level? We can set emission limits at a national level.

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+            if granularity == "node":
+                atmosphere = f"{node} {state_sector} atmosphere"
+            else:  # state
+                atmosphere = f"{state} {state_sector} atmosphere"
+            buses_atmosphere_all.append(atmosphere)
+            if atmosphere not in exists_atmosphere:
+                buses_atmosphere_unique.append(atmosphere)
+                buses_co2_account.append(bus2)
+                exists_atmosphere.add(atmosphere)
+            dac = f"{node} {state_sector} dac"
```
trevorb1, line 1163: Im not sure Im following the state/sector representation? Shouldnt dac be independent of sector? Do you have an image with multiple sectors that describes what happens? Cause this makes it look like it happens at a sector level? 

image

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+                buses_ac.append(node)
+                links_dac.append(dac)
+                exists_dac.add(dac)
+
+        # add node or state level buses to represent (air) atmosphere where CO2 emissions are sent to (on a per sector basis)
+        n.madd(
+            "Bus",
+            buses_atmosphere_unique,
+            carrier="co2",
+        )
```
trevorb1, line 1176: I think we also need to update the current emissions tracking, so co2 can be removed from earlier timesteps? Like right now once emissions are released, if they arn't removed in that timestep, they have be sent to atmosphere and left there due to the one-diretional store?

trevorb1: I guess maybe we set `e_min_pu` to `-1` and `e_min` to `0`? But Im not sure off the top of my head if that has any other implications?

workflow/scripts/add_extra_components.py
```diff
@@ -882,6 +883,346 @@ def trim_network(n, trim_topology):
...
+        for index in links.index:
+            bus2 = links.loc[index]["bus2"]  # e.g. "CA pwr-co2"
+            node = " ".join(index.split(" ")[:elements])  # e.g. "p9" if ReEDS or "p100 0" if TAMU
+            state = bus2.split(" ")[0]  # e.g. "CA"
+            state_sector = bus2.split(" ")[1].split("-")[0]  # e.g. "pwr"
+            if granularity == "node":
+                atmosphere = f"{node} {state_sector} atmosphere"
+            else:  # state
+                atmosphere = f"{state} {state_sector} atmosphere"
+            buses_atmosphere_all.append(atmosphere)
```
trevorb1, line 1158: Sorry, Im not really understanding how this interfaces with the current emission tracking in sector? Like why cant we do national level?

workflow/scripts/add_extra_components.py
trevorb1: image

workflow/scripts/add_extra_components.py
ktehranchi: hmmm..... I'm fine with it the way it is just to reduce the number of carriers.... in the future if we have more complex Co2 Capture/Storage systems then it makes more sense? No strong opinions here

trevorb1: Fair point! Thanks!

workflow/repo_data/config/config.default.yaml
```diff
@@ -92,13 +92,31 @@ lines:
...
 
+# docs :
+co2:
+  storage: false   # [false, true]
+  network:
+    enable: false   # [false, true]
+    capital_cost: 2736000   # in USD/12-inch diameter/mile (cost based on EPA estimation, which assumes a 12-inch diameter and a cost of $228000 per inch for each mile of CO2 pipeline built)
+    marginal_cost: 4   # in USD/tCO2 (average cost based on NETL estimation, which ranges between $2 and $6 to transport one tonne of captured CO2)
+    lifetime: 40   # in years (average lifetime based on NETL estimation, which ranges between 30 and 50 years)
+    discount_rate: 0.07   # in percentage
```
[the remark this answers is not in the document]
trevorb1: But this discount rate is supposed to represent `7%` ? `7%` = `0.07` `per_unit`? Maybe I am just getting confused with terminology tho 😅

workflow/repo_data/config/config.default.yaml
```diff
@@ -92,13 +92,31 @@ lines:
...
+
+# docs :
+dac:
+  enable: false   # [false, true]
+  granularity: "node"   # ["node", "state", "nation"] (note: "nation" is not applicable for a network based on sectors - it defaults to "node" in this case)
+  capital_cost: 6000000   # in USD/tCO2/h (cost based on Danish Energy Agency estimation to capture one tonne of CO2 per hour)   # TODO: check this and adapt it to the USA context
+  electricity_input: 2.5   # in MWh/tCO2 (average consumption based on https://doi.org/10.1557/s43581-024-00091-5, which ranges between 2 to 3 MWh to capture one tonne of CO2 using electricity only - no heat)
+  lifetime: 20   # in years
+  discount_rate: 0.07   # in percentage
+
```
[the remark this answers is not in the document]
ricnogfer: This review/feedback is solved/implmented with commit https://github.com/ricnogfer/pypsa-usa/commit/a9673f8c96f75defbf63509aa0fcfa2efc47f686.

workflow/scripts/add_extra_components.py
ricnogfer: This review/feedback is solved/implemented with commit https://github.com/ricnogfer/pypsa-usa/commit/3854dfaf85bad703b949998ae4b75aa43c64cccd.

workflow/scripts/add_extra_components.py
ricnogfer: This review/feedback is solved/implemented with commit https://github.com/ricnogfer/pypsa-usa/commit/ca6e8dab1117e1c222f52bbcc9a1d0683ea8011e.

workflow/scripts/add_extra_components.py
ricnogfer: Need to check this particular suggestion before making a change as ``links`` is not a scalar.

workflow/scripts/add_extra_components.py
ricnogfer: Thanks for the feedback @trevorb1!

Would it be fine on your side if we have just one bus representing the national atmosphere and then have DAC links (one for each node composing the network) capturing CO2 for this bus and then sending the captured CO2 to the respective capture bus (for this CO2 to then be either stored underground or transported to another node)? If you agree, I could change the logic to do this. Please confirm - thanks.

workflow/scripts/add_extra_components.py
ricnogfer: For the sector-based CO2 tracking system (e.g. `CA pwr-co2` for California's power sector) to function correctly, the bus representing the atmosphere must be differentiated by sector - whether at the node or state level, as shown in the two images above. If all sectors emit CO2 into a single bus atmosphere and DAC is not used (due to not being cost-effective), it becomes impossible to trace emissions back to their originating sectors. This undermines the entire purpose of the CO2 tracking system.

@trevorb1: If there is an alternative way to represent a node or state atmosphere bus without sector discrimination while still preserving the functionality of the CO2 tracking system, I would be happy to revise the logic accordingly.

workflow/repo_data/config/config.default.yaml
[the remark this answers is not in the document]
ktehranchi: you're right!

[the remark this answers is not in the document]
ricnogfer: This review/feedback is solved/implemented with commit https://github.com/ricnogfer/pypsa-usa/commit/76468a213b1ed7f1c108e7523830b7fe0e14187c.

workflow/scripts/add_extra_components.py
ricnogfer: For DAC to work at a national level means that only one single bus should be created to represent the (nation) atmosphere (currently, each state has its own bus representing its atmosphere). The problem arises if DAC is not cost-effective: emitted CO2 into the atmosphere will end up being stored across the many stores (all attached to this single bus atmosphere) that are meant to account for the CO2 emitted by states and sectors. This will  make the sector-based CO2 tracking system no longer able to correctly "tell" where the emitted CO2 came from, state or sector-wise.

That said, if we really want to have DAC working at a national level, the sector-based CO2 tracking system should no longer be discriminated by state as it is currently and only by sector (if it should not be discriminated by sector then we fall on another issue described here https://github.com/PyPSA/pypsa-usa/pull/622#discussion_r2254729185).

@trevorb1: If there is an alternative way to represent a node or state atmosphere bus without sector discrimination while still preserving the functionality of the CO2 tracking system, I would be happy to revise the logic accordingly!

workflow/scripts/add_extra_components.py
ricnogfer: Please refer to https://github.com/PyPSA/pypsa-usa/pull/622#discussion_r2264684779 for additional details.

workflow/scripts/add_extra_components.py
ricnogfer: This review/feedback is solved/implemented with commit https://github.com/ricnogfer/pypsa-usa/commit/319d04b61b60363ba06437840699ca8023ef9a99.

workflow/scripts/add_extra_components.py
ricnogfer: Hi @trevorb1,

Would you mind confirming this? Many thanks in advance!

workflow/scripts/add_extra_components.py
trevorb1: Great, @ricnogfer! Apologies for using the wrong syntax!

trevorb1:
@ricnogfer, so sorry for the slow reply in addressing your answers! 

I think we should do as @ktehranchi originally suggested, and not allow DAC for sector studies right now (see my code suggestion, but please feel free to update if its better somewhere else). Instead, we should move the issues with implementing it to separate tickets. In my opinion, installing DAC at a total model scope level should be the default, which isnt available right now (as you describe [here](https://github.com/PyPSA/pypsa-usa/pull/622#discussion_r2254729185)). 

The solution may be any one of the following: 
- Create a central Co2 atmosphere store
- Redo how sector emission tracking is done. So instead of tracking by individual sectors, we just process activity after the solve to see how much emissions are released per sector. (Probably this is the best options?) 
- Link DAC up to all stores and constrain it so it pulls emissions from each store

Moreover, may be worth it to see how PyPSA-Eur does it. I originally set it up to match my research questions. As I don't have a need for DAC, I wont have time to address this. 

@ricnogfer are you okay if we block DAC in sector, move this conversation to new issue ticket(s), and then get this merged so it can be used for non-sector studies?

ricnogfer:
> @ricnogfer, so sorry for the slow reply in addressing your answers!
> 
> I think we should do as @ktehranchi originally suggested, and not allow DAC for sector studies right now (see my code suggestion, but please feel free to update if its better somewhere else). Instead, we should move the issues with implementing it to separate ticket (see ticket #652). As we are comfortable with the functionality for non-sector systems, it would be good to get that merged!
> 
> My biggest hold up with DAC for sector is that it can't be installed at a total model scope. This is due to the issues you describe [here](https://github.com/PyPSA/pypsa-usa/pull/622#discussion_r2254729185). If there is a use case where DAC is needed for sector, I wont hold this PR back tho!
> 
> The solution may be any one of the following:
> 
>     * Create a central Co2 atmosphere store (similar to how power sector is implemented?)
> 
>     * Redo how sector emission tracking is done. So instead of tracking by individual sectors, we just process activity after the solve to see how much emissions are released per sector. (Probably this is the best options?)
> 
>     * Link DAC up to all stores and constrain it so it pulls emissions from each store
> 
> 
> Moreover, may be worth it to see how PyPSA-Eur does it. I originally set it up to match my research questions. As I don't have a need for DAC, I wont have time to address this.
> 
> @ricnogfer are you okay if we block DAC in sector, move this conversation to new issue ticket(s), and then get this merged so it can be used for non-sector studies?


@trevorb1: All good to block DAC functionality for sector-based networks if you wish. Also, I believe option #2 (detailed in ticket https://github.com/PyPSA/pypsa-usa/issues/652) to be the best approach for moving forward with this functionality when the opportunity arises!

workflow/scripts/add_sectors.py
```diff
@@ -753,4 +754,27 @@ def get_pwr_co2_intensity(carrier: str, costs: pd.DataFrame) -> float:
...
+            )
+
+    # add node level DAC capabilities
+    if snakemake.config["dac"]["enable"]:
+        if snakemake.config["co2"]["storage"]:
+            logger.info("Adding node level DAC capabilities")
+            add_dac(n, snakemake.config, True)
+        else:
+            logger.warning("Not adding node level DAC capabilities given that CO2 (underground) storage is not enabled")
+
```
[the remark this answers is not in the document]
ricnogfer: This review/feedback is solved/implemented with commit https://github.com/PyPSA/pypsa-usa/pull/622/commits/d15d5262b1587b44e658ab0fd06bc1d47d105f24.

workflow/scripts/add_extra_components.py
trevorb1: See item 5 in issue #652

trevorb1 (approved):

PR #627, opened 2025-06-25
WUS Integration

mchjones:
### Changes and their purpose
Changes made to 
- build_renewable_profiles.py: if cf_source is set to WUS, it will substitute in the WUS capacity factor timeseries for wind and solar for the closest lat/lon into the profile for all the buses in the system after the it does the land use calculations, etc. using the ERA5 data/atlite
- build_electricity.smk: changed calls for build_renewable_profiles to have it generate multiple profiles and list the capacity factor source in the generated profile name, changed calls for add_electricity so it calls the multiple profiles
- add_electricity.py: changed the code where it puts the renewable profiles in and just copies over the same profile year for each planning horizon - this concatenates the profiles generated (in order) and attaches them to the buses in the network
- add_demand.py, add_fuel_prices.py: change the way snapshots are called because of the new way that snapshots are defined in the config file (see below)

### Config File Changes
I'm not sure if it updates or shows the config files but just in case:

In your run's config file (i.e. the config.default.yaml if that's where you're starting from), the renewable weather year information needs to look like this:

renewable_weather_years: [2012,2019,2023] # need to be historical

snapshots:
  '2019':
    start: "2019-01-01"
    end: "2020-01-01"
    inclusive: "left"
  '2012':
    start: "2012-01-01"
    end: "2013-01-01"
    inclusive: "left"
  '2023':
    start: "2023-01-01"
    end: "2024-01-01"
    inclusive: "left"

And in the config.common.yaml, you need to add:

cf_source: "WUS" # WUS or ERA5
cf_path: "T:/WRFDownscaled/ec-earth3-veg_r1i1p1f1_ssp370_bc/Annual_Solar_Wind" 

If you're looking to use the WUS data, let me know and I can help facilitate getting you access to the capacity factor data; you'll need to change the cf_path to wherever you store the data. If you just want to use the multiple weather profile capability, just set cf_source equal to ERA5 and it will use the ERA5 weather data for the years you specify for the renewable weather years, no need to worry about the cf_path.


ktehranchi:
Hey Megan- thanks for opening this PR! I noticed you originally were merging into `develop` then switched to `wus-integration` branch instead. It looks like your diff is quiet large from `develop` because you are working off a fork of lyssa's fork.

We'll need to change the base to `develop` and perform a `rebase` operation on your current branch, which means we'll need to take all the changes you made downstream of [this commit](https://github.com/PyPSA/pypsa-usa/pull/627/commits/cfa4ed42307d2563a8de49cfea5fd5ab4837e623) and apply them to the latest develop branch. sometime you could just use the `rebase` operation directly, but because this is a large diff including lyssa's work that was already merged into `develop`, I would recommend creating a new branch from `develop` and directly moving your commits to the 5 files you mentioned.

I know these git tasks can get annoying/daunting at first, so I would be happy to jump on a call to help on this rebase/PR. In the long-run I recommend keeping your active working branch up-to-date with the latest `develop` branch to ensure you can take advantage of new bug-fixes and features. I'm still refining/learning my research GitHub management routines all the time, but I'm happy to share my workflows.

mchjones:
Ah yeah, I wasn't sure which branch I should be merging into, haha. I will make an attempt at doing as you outlined; if I run into issues, I will let you know and we can set up a meeting. I'll keep my work up-to-date with develop instead of master in the future!

Thank you for your help!

PR #631, opened 2025-07-01
WUS Integration [NEW]

mchjones:
I closed the previous pull request and made a new one (I couldn't figure out how to change the branch I was merging into develop) - I made a local branch from develop, then cherry-picked the five wus-integration commits as suggested. Hopefully it works this time? Let me know if there's anything I need to fix!

### Changes and their purpose
Changes made to 
- build_renewable_profiles.py: if cf_source is set to WUS, it will substitute in the WUS capacity factor timeseries for wind and solar for the closest lat/lon into the profile for all the buses in the system after the it does the land use calculations, etc. using the ERA5 data/atlite
- build_electricity.smk: changed calls for build_renewable_profiles to have it generate multiple profiles and list the capacity factor source in the generated profile name, changed calls for add_electricity so it calls the multiple profiles
- add_electricity.py: changed the code where it puts the renewable profiles in and just copies over the same profile year for each planning horizon - this concatenates the profiles generated (in order) and attaches them to the buses in the network
- add_demand.py, add_fuel_prices.py: change the way snapshots are called because of the new way that snapshots are defined in the config file (see below)

There are also changes to retrieve_databundles.py to facilitate zipfile downloads on windows, that has nothing to do with the WUS stuff.

### Config File Changes
I'm not sure if it updates or shows the config files but just in case:

In your run's config file (i.e. the config.default.yaml if that's where you're starting from), the renewable weather year information needs to look like this:
```
renewable_weather_years: [2012,2019,2023] # need to be historical

snapshots:
  '2019':
    start: "2019-01-01"
    end: "2020-01-01"
    inclusive: "left"
  '2012':
    start: "2012-01-01"
    end: "2013-01-01"
    inclusive: "left"
  '2023':
    start: "2023-01-01"
    end: "2024-01-01"
    inclusive: "left"
```
And in the config.common.yaml, you need to add:

```
cf_source: "WUS" # WUS or ERA5
cf_path: "T:/WRFDownscaled/ec-earth3-veg_r1i1p1f1_ssp370_bc/Annual_Solar_Wind" 
```

If you're looking to use the WUS data, let me know and I can help facilitate getting you access to the capacity factor data; you'll need to change the cf_path to wherever you store the data. If you just want to use the multiple weather profile capability, just set cf_source equal to ERA5 and it will use the ERA5 weather data for the years you specify for the renewable weather years, no need to worry about the cf_path.

### DIFF
--- a/workflow/repo_data/config/config.common.yaml
+++ b/workflow/repo_data/config/config.common.yaml
@@ -1,5 +1,8 @@
 pudl_path: s3://pudl.catalyst.coop/v2025.5.0
 
+cf_source: "ERA5" # WUS or ERA5
+cf_path: "T:/WRFDownscaled/taiesm1_r1i1p1f1_ssp370_bc/Annual_Solar_Wind" # set to local folder with capacity factor data (contact mchjones@umich.edu about WUS data)
+
 # docs :
 renewable:
   EGS:
--- a/workflow/repo_data/config/config.default.yaml
+++ b/workflow/repo_data/config/config.default.yaml
@@ -34,12 +34,21 @@ model_topology:
 enable:
   build_cutout: false
 
-renewable_weather_years: [2019]
+renewable_weather_years: [2012,2012,2012] # need to be historical
 
 snapshots:
-  start: "2019-01-01"
-  end: "2020-01-01"
-  inclusive: "left"
+  '2019':
+    start: "2019-01-01"
+    end: "2020-01-01"
+    inclusive: "left"
+  '2012':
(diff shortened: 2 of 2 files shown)


ktehranchi:
LGTM! looks like we need to resolve a bunch of merge conflicts first. Looks like the force push trevor did 3 weeks ago caused the same issue with the large diff. Ugh these diffs are annoying to resolve. I can do it since you already did it last time lol.

I can make a copy of your branch and open a new PR after everything is resolved. Sorry about that!

PR #648, opened 2025-08-06
Bi-directional links

### LINKED ISSUE #647: Unexpected behavior of bi-directional link with non-unit efficiency
### Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-usa/tree/master) branch
- [x] I am running on an up-to-date [`pypsa-usa` environment](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

### The Issue

The implementation of link efficiency of PyPSA is $p_1 = p_0 * efficiency$. This works fine when the power flow is uni-directional. For example, if the efficiency is 0.9, bus1 will receive 0.9 MW while bus0 sends 1MW. However, when the link is bi-directional, and the power is negative (say -1MW), the following could happen: bus0 is receving 1MW but bus1 is only sending 0.9MW, which is obviously what not we want and will severely affect the model's behavior - the link is actually generating power from nothing! Following is an example:

We are modeling transmission as bi-directional link with loss:
Image
If we look at the power flow, n.links_t.p0:
Image
and n.links_t.p1:
Image
we can see that in the first four timesteps, bus0 (AL) is receiving more power from bus1 (TN) than bus1 (TN) is actually sending.

As a result, I think we should model all lossy bi-directional link separately as one forward link and one backward link.

Any thoughts? @ktehranchi @trevorb1 @mchjones 

### Steps To Reproduce

_No response_

### Expected Behavior

_No response_

### Error Message

```shell

```

### Anything else?

_No response_

WeiAi-Energy:
Closes #647

Added efficiency loss for HVDC transmission according to the assumptions in https://www.nature.com/articles/s41560-025-01752-6#Sec14. Modeled lossy bi-directional links as _fwd and _rev uni-directional link pairs and added associated constraints in `solve_network.py`.

## Checklist


ktehranchi (approved):
Looks good to me! Thanks for updating the WACC for links also- I had that on my local branch but hadn't pushed yet!

WeiAi-Energy:
@ktehranchi Sorry in the previous version I coded in a way that the capital costs are calculated twice! Now fixed it. I also realized we can simply set existing AC links as extendable instead of adding extra AC_exp links. The constraints can also be simplified.

trevorb1 (approved):
This is awesome! Way to go @WeiAi-Energy 🎉🎉🎉🎉

PR #665, opened 2025-08-20
Sync `develop` with `master` `v0.7.0`

trevorb1:
I think we need to PR this as the develop branch is protected so we cant directly `rebase` or `reset` the develop branch.


trevorb1:
We just force reset instead

PR #655, opened 2025-08-18
Electrical Import/Export Functionality 

trevorb1:
Closes #640, #617

## Changes proposed in this Pull Request

PR #615 introduced the `trim` feature to allow imports from outside the model scope. This PR build on this to generalize the import/export functionality. Specifically, it allows user to: 

1. Allow import and exports independently
2. Apply wholesale electrical prices to the imports/exports 
3. Constrain capacity (if desired) of import/export links 
4. Apply maximum import/export volumes over different balancing periods 
5. Apply custom emission factors to the imports 

Some notes on how this is different from the `trim` feature. 
- Trim removes portions of the network. This import/export configuration does not. Instead, if finds the connections from the `flowgates` data to regions outside of the network and adds in the trading.
- Trade is represented by `links` and `stores`, unlike the `trim` feature which uses generators 
- Import costs and limits can be configured
- Works with sector studies

Some other notes: 
- I spent a bunch of time trying to pull in the `eia930` data, which gives net interchange by hour between balancing authorities. I ended up abandoning this though as I couldn't get reliable validation results when trying to manually match BAs to the different zones (ie. the trade volume ended up being quite sensitive in my (very crude!) tests) 
- All functionality in here is a superset of the the `trim` functionality. ie. you can replicate the `trim` functionality with this and selecting the model topology in the wildcards. 

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Changed dependencies are added to `envs/environment.yaml`.
- [x] Changes in configuration options are added in all of `config.default.yaml`.
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv`.

Issue #640, opened 2025-07-23
Trim feature not compatible with sector studies


trevorb1:
### Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-usa/tree/master) branch
- [x] I am running on an up-to-date [`pypsa-usa` environment](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

### The Issue

PR #615 introduces the trim feature to capture imports from other regions outside of the model scope. This is not currently compatible with sector studies as trip will add states that doe not exists (for example `imports_CA`). This is an issue as sector studies build natural resource foundations and track natural gas usage at a state level using `reeds_state`. 

### Error Message

```shell

```

### Anything else?

Update the filtering on how these foundations are created within the `add_sectors` script by excluding any imports/exports prefix.

PR #668, opened 2025-08-21
(v2) Electrical Import/Export Functionality

### LINKED ISSUE #617: Electrical connections to regions outside model scope
### Feature Request

Added by @ktehranchi in PR #615 is the `trim` feature which retains "all buses one-step outside of the network". The goal of this is to reduce artifacting of runs where we model only a single state (or similar small spatial scopes). To build on this, I am wondering if we can remove generators at the 'trimmed' buses, and instead allow imports/exports up to certain limits based on historical data. 

### Suggested Solution

In [this PR](https://github.com/DeltaE/pypsa-gsa/pull/32/), I have (crudely) implemented this at an ISO level. It should be able to be downscaled to state level though, given the available data. What this PR includes is: 

- Pulls in data on historical net-flows, and balance over one of month. The data I use is "core_eia930__hourly_interchange" from PUDL [here](https://viewer.catalyst.coop/). I guess this may difficult to generalize though, since this data does not come in at BA level. Moreover, this is only net, so Im not sure how sensitive it is to the choice of balancing period (which can go down to hourly).
- I retain the capacity limits at a BA level into neighbouring ISOs. This is same as what 'trim' retains, I believe! (But my implementation isn't integrated into PyPSA-USA yet, so the current one is better!).
- Pull in historical [EIA wholesale electricity market data](https://www.eia.gov/electricity/wholesale/) as the import/export price.

One key difference between this implementation and the 'trim' (currently implemented)  is that emissions associated with the trimmed buses are no longer counted against state level emission budgets, but emissions associated with electricity generated for exports is accounted for.

### Additional Info

Additionally, the [EIA has state level electricity imports](https://www.eia.gov/opendata/browser/electricity/state-electricity-profiles/source-disposition?frequency=annual&data=net-interstate-trade;&sortColumn=period;&sortDirection=desc;), which may be easier to parse than the straight 930 data. Moreover, the [EIA provides prices at state level](https://www.eia.gov/opendata/browser/electricity/retail-sales?frequency=monthly&data=price;&facets=sectorid;&sectorid=ALL;&start=2024-01&end=2024-12&sortColumn=period;&sortDirection=desc;), instead of just at hubs. This is end use retail sales instead of wholesale price tho.

trevorb1:
Closes #617 #640

Supersedes PR #655 to clean up commit history. No logic changes between PRs. 

## Checklist

### DIFF
--- a/docs/source/configtables/electricity.csv
+++ b/docs/source/configtables/electricity.csv
@@ -28,7 +28,7 @@ demand_response:,,,Settings to activate and configure demand response
 ,,,
 imports:,,,Configure electric imports from regions outside of model scope
 -enable,,``true`` or ``false``,Enable electric imports
--costs,$/MWh,``true`` or ``false`` or ``float`` of ``carrier``,Cost of electric imports from regions outside of model scope. ``False`` will set cost to ``0``. ``True`` will pull in wholesales electric prices. ``float`` will assign a user specified value. ``carrier`` will take average marginal cost of the carrier..
+-costs,$/MWh,``wholesale`` or ``<name of carrier>`` or ``float``,Cost of electric imports from regions outside of model scope. ``Wholesale`` will use in monthly wholesales electric prices. ``float`` will assign a user specified value. ``carrier`` will take average marginal cost of the carrier.
 -co2_emissions,CO2/MWh,``float``,CO2 emissions of electric imports from regions outside of model scope.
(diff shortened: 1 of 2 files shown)


trevorb1:
This def works on power sector! Just need to run a few more tests on sector coupled networks

trevorb1:
Actually, need to double check that RPS still functions correctly with imports

PR #687, opened 2025-09-22
[fixed] WUS Integration

ktehranchi:
Resolves conflicts from #631


ktehranchi:
Outstanding issues with this:

1. The weather_year wildcard causes a cutout to be repeatedly downloaded if the weather year is the same per planning horizon ie [2012, 2012, 2012]
2. Reconciliation with @asiazzzhang godeeep method for tracking renewable profile years. the configuration file is messy and things need to be combined
3. The fix for the EFS zipdeflate works on Mac (by bypassing it) but should be tested on windows

ktehranchi:
@asiazzzhang can we close this? did you end up implementing this changes in your recent PR?

asiazzzhang:
@ktehranchi ok! I didn't integrate the WUS dataset though, not sure if that was a separate PR that was already closed.

PR #699, opened 2025-10-09
[WIP] Major Refactor to enable python api access and packaging 

ktehranchi:
This commit enables users to `pip install pypsa-usa` and run workflows from a python api as described in the new quick start guide below.

None of the python files were changed nor the snakemake rules besides the paths.... I will request reviews from folks soon

### Option 1: PyPI Installation (Recommended)

```bash
pip install pypsa-usa
```

### Option 2: Development Installation

```bash
git clone https://github.com/PyPSA/pypsa-usa.git
cd pypsa-usa
uv venv && source .venv/bin/activate
uv pip install -e .
```

### Using the Python API

```python
from pypsa_usa.api import run_workflow, set_default_workspace
from pathlib import Path

# Set your default workspace (only needed once)
set_default_workspace("/path/to/my/project/workspace")

# Run a simple workflow with default settings
success = run_workflow(cores=4)

if success:
    print("Workflow completed successfully!")
    print("Results are in your workspace results/ directory")
else:
    print("Workflow failed. Check logs/ for details.")
```

### Using Snakemake Directly (Developmer Installation)

```bash
# Navigate to workflow directory
cd workflow

# Run with default configuration
uv run snakemake -j4 --configfile config/config.default.yaml
```


src/pypsa_usa/api/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+    # Get template path
+    template_path = get_workflow_path() / "config" / template
+
+    # Get user config path
+    user_config_dir = workspace_path / "config"
+    user_config_dir.mkdir(parents=True, exist_ok=True)
+    user_config_path = user_config_dir / config_name
+
+    # Copy template to user config
+    shutil.copy2(template_path, user_config_path)
```
trevorb1: Im a little confused on why we are copying configs?

trevorb1 (commented):
Thanks for starting this, @ktehranchi! I think this is super cool! I totally agree with @mchjones comment of: 

>good functionality to develop and support to make the tool more accessible and practical for use in an educational setting

Unfortunately, I wasn't able to get this running. Not sure if I am just misunderstanding what path to pass to `set_default_workspace()`? Before I leave too many comments on the PR, I will wait to see if others have had similar running issues!

trevorb1:
Alternatively, we could also look at containerizing this via docker. This might be the best of both worlds. It is flexible enough so that both developers and general users would install it (as it allows for pure snakemake calls still). Plus, that way if any commands come up that are needed (ie. `--unlock` or `--rerun-incomplete`), they are easy enough to append to the docker call. I believe pypsa-eur has containerized their workflow [here](https://github.com/PyPSA/pypsa-eur/blob/master/docker/dev-env/Dockerfile). 

Downsides are it might be a bit of a pain to setup and manage. Ive never done it before for a real project, so idk the types of issues that are likely to come up. May be worth checking in with the pypsa-eur crew if we want to explore this more to get their thoughts. Cause it may also just be more hassle then its worth, even for developers to use.

PR #753, opened 2026-05-27
fix(build_bus_regions): dissolve duplicate bus rows after empty-county assignment

### LINKED ISSUE #752: bug: build_renewable_profiles crashes with ValueError when regions_onshore has duplicate bus names
### Environment
- **OS**: Windows 11
- **Interconnect**: western
- **Rule**: `build_renewable_profiles` (technology: `onwind`)

### Description

`build_renewable_profiles` crashes with a `ValueError` during the "Calculate average distances" step when `regions_onshore.geojson` contains duplicate bus names.

```
INFO:__main__:Calculate average distances.
Traceback (most recent call last):
  File "...build_renewable_profiles.py", line 264, in <module>
    co = coords[nz_b]
ValueError: Item wrong length 3 instead of 18360.
```

### Root Cause

In `build_bus_regions.py`, when empty counties (counties within the interconnect footprint but containing no substations) are assigned to the nearest bus, a new GeoDataFrame row is appended that reuses the same `name` as the existing bus entry:

```python
empty_region_rows.append({
    "name": nearest_sub["name"],  # same name as an existing bus!
    ...
    "geometry": county_geom,
})
```

If two or more empty counties share the same nearest bus, that bus gets **2+ duplicate `name` entries** in `regions_onshore.geojson`.

Downstream in `build_renewable_profiles.py`:

```python
regions = regions.set_index("name").rename_axis("bus")  # duplicate index
...
layoutmatrix = (layout * availability).stack(spatial=["y", "x"])
for bus in buses:
    row = layoutmatrix.sel(bus=bus).data   # returns (k, n_spatial) if bus appears k times
    nz_b = row != 0                        # shape (k, n_spatial) — 2-D!
    co = coords[nz_b]                      # ValueError: length k != n_spatial
```

When `sel(bus=bus)` finds `k` duplicate entries it returns a `(k x n_spatial)` array. Pandas then treats `nz_b` as a length-`k` boolean column selector, raising `ValueError: Item wrong length k instead of n_spatial`.

### Steps to Reproduce

Run `snakemake` with the `western` interconnect (or any interconnect with counties that contain no substations):

```
snakemake -j1 resources/Default/western/profile_onwind.nc
```

### Expected Behavior

The rule completes successfully, producing `profile_onwind.nc`.

### Fix

After concatenating the empty-county rows, dissolve the GeoDataFrame by `name`, taking the **union of all geometries** that share a bus name. Bus `x`/`y` coordinates are preserved from the first (original) entry using `aggfunc="first"`.

A PR with this fix is attached.

WeiAi-Energy:
Fixes #752

## Summary

`build_renewable_profiles` crashed with:

```
ValueError: Item wrong length 3 instead of 18360.
```

during the "Calculate average distances" step when `regions_onshore.geojson` contained duplicate bus names.

## Root Cause

In `build_bus_regions.py`, empty counties (counties within the interconnect footprint but with no substations) are assigned to the nearest bus by appending a new GeoDataFrame row that reuses the same `name` as the existing bus entry. When two or more empty counties share the same nearest bus, that bus acquires 2+ duplicate rows in `regions_onshore.geojson`.

Downstream in `build_renewable_profiles.py`:

```python
row = layoutmatrix.sel(bus=bus).data   # returns shape (k, n_spatial) for k duplicates
nz_b = row != 0                        # 2-D boolean mask
co = coords[nz_b]                      # ValueError: length k != n_spatial
```

`xr.DataArray.sel()` returns all matching entries when the coordinate label is duplicated, producing a 2-D array instead of the expected 1-D spatial vector. Pandas then misinterprets the 2-D boolean mask as a column selector of length `k`, raising `ValueError: Item wrong length k instead of n_spatial`.

## Fix

After concatenating the empty-county rows into `onshore_regions_concat`, call `GeoDataFrame.dissolve(by="name")` to merge all rows that share the same bus name into a single row with the union of their geometries:

```python
onshore_regions_concat = onshore_regions_concat.dissolve(
    by="name",
    aggfunc={"x": "first", "y": "first", "country": "first"},
).reset_index()
```

- **Geometry**: union of the original bus Voronoi cell + all empty-county polygons assigned to it — physically correct, as the bus now represents that larger area.
- **x / y**: preserved from the first occurrence in concat order, which is always the original substation location (empty-county rows are appended after the original regions).
- **No-op when there are no empty counties**: the `if empty_counties:` guard prevents the dissolve from running unless duplicates were actually introduced.

## Testing

Unit test confirms three properties:
1. Bug is reproducible without the fix (`ValueError` raised).
2. With the fix, the downstream loop completes without error.
3. Bus `x`/`y` coordinates are unchanged after dissolve.
4. When there are no duplicates the dissolve is a no-op.


WeiAi-Energy:
Closing: the identical fix (dissolve by name, aggfunc first) has already been applied in the \develop\ branch (same block, same logic, plus explicit \set_crs\ restoration). This PR is therefore superseded by that change and would conflict. The bug will be resolved for \master\ users when \develop\ is next merged.

PR #773, opened 2026-08-12
fix(rps): credit behind-the-meter rooftop solar against RPS targets

WilsonHM18:
Rooftop solar (solar-rooftop) was excluded from RPS-eligible carriers and its generation was embedded in net-load data, causing the RPS constraint to be over-tightened.

Changes:
- Add "solar-rooftop" to RPS_CARRIERS so BTM generation counts toward the portfolio standard
- Fix a return→continue bug that silently skipped states with no matching portfolio-standard row
- Implement BTM credit adjustment: rhs = pct*net_load - (1-pct)*btm_rooftop so that rooftop generation already embedded in load is not double-penalized
- Add retrieve_small_scale_solar rule (EIA API v2 with bundled CSV fallback) and wire small_scale_solar.csv into solve_network inputs
- Add rec_trading_zone to test network fixtures, restore sector=False parameter for call-site compatibility, and add test_btm_solar_credit_reduces_rps_rhs

Full-pipeline test confirmed:
  net_load=97.2 TWh, btm_rooftop=39.80 TWh, adjusted_rhs=21.5 TWh
  Solver: optimal

## Checklist

- [X ] I tested my contribution locally and it seems to work fine.
- [X] Code and workflow changes are sufficiently documented.
- [NA] Changed dependencies are added to `envs/environment.yaml`.
- [NA] Changes in configuration options are added in all of `config.default.yaml`.
- [NA] Changes in configuration options are also documented in `doc/configtables/*.csv`.


workflow/scripts/retrieve_small_scale_solar.py
```diff
@@ -0,0 +1,214 @@
...
+logger = logging.getLogger(__name__)
+
+# EIA API v2 endpoint for electric power operational data
+_EIA_URL = "https://api.eia.gov/v2/electricity/electric-power-operational-data/data/"
+
+# Sector 98 = small-scale solar (below 1 MW nameplate)
+_SECTOR_ID = "98"
+_FUEL_TYPE = "SUN"
+
+# All CONUS state abbreviations covered by pypsa-usa
```
trevorb1: Can we reuse the `STATE_2_CODE` or `CODE_2_STATE` constant [here](https://github.com/PyPSA/pypsa-usa/blob/fe2c82953ed3123ddfd1eb7475fadd5abf0d2880/workflow/scripts/constants.py#L486-L563) for this?

workflow/scripts/retrieve_small_scale_solar.py
```diff
@@ -0,0 +1,214 @@
...
+
+def _load_fallback(fallback_path: str) -> pd.DataFrame:
+    """Load a pre-bundled CSV when no EIA API key is available."""
+    logger.warning(
+        "No EIA API key provided. Loading bundled small-scale solar data from "
+        f"{fallback_path}. This data may not match your planning horizons exactly; "
+        "the most recent available year will be used for future years.",
+    )
+    df = pd.read_csv(fallback_path, dtype={"state": str, "year": int, "generation_mwh": float})
+    return df
```
trevorb1: There is a [dedicated EIA module](https://github.com/PyPSA/pypsa-usa/blob/master/workflow/scripts/eia.py) for interfacing with its API. Would be great to consolidate EIA API logic there. It is set up so you should just be able to create a new [`DataExtractor` instance](https://github.com/PyPSA/pypsa-usa/blob/fe2c82953ed3123ddfd1eb7475fadd5abf0d2880/workflow/scripts/eia.py#L494-L523) and call it under the [`Production` class](https://github.com/PyPSA/pypsa-usa/blob/fe2c82953ed3123ddfd1eb7475fadd5abf0d2880/workflow/scripts/eia.py#L257-L258). If the EIA module logic is not clear though, please let me know!

workflow/scripts/retrieve_small_scale_solar.py
```diff
@@ -0,0 +1,214 @@
...
+    api_key = snakemake.params.get("eia_api", None)
+    planning_horizons = snakemake.params.planning_horizons
+    fallback_path = snakemake.input.fallback
+
+    if api_key:
+        # Fetch years from the earliest data year through the last planning horizon.
+        # EIA small-scale solar data starts around 2014.
+        start_year = 2014
+        end_year = max(planning_horizons)
+        df = _fetch_from_eia(api_key, start_year, end_year)
```
trevorb1: Probably replace with a call to the EIA module with something along the lines of: 

```python
import eia

production = eia.Production("rooftop_solar", "", 2014, api_key)
df = production.get_data()
``` 

But, also open to suggestions!

workflow/scripts/retrieve_small_scale_solar.py
trevorb1: To get multiple years you will need to instantiate the class multiple times, which I guess isn't fantastic. Alternatively, maybe implementing something similar to [how we get multiple years from the AEO](https://github.com/PyPSA/pypsa-usa/blob/fe2c82953ed3123ddfd1eb7475fadd5abf0d2880/workflow/scripts/eia.py#L289-L303) would work here?

workflow/scripts/opts/policy.py
```diff
@@ -247,18 +248,44 @@ def _collapse_portfolio_standards(n: pypsa.Network, planning_horizons: list[int]
...
+    each constraint is adjusted from net load to gross load by adding back the
+    behind-the-meter (rooftop) solar generation that is embedded as a demand
+    reduction in the EIA 930 input data.  This ensures that existing rooftop
+    solar receives credit toward the RPS target even when it is not explicitly
+    modelled as a Generator in the network.
+
+    The adjusted RHS is:
+        rhs = pct * net_load - (1 - pct) * rooftop_gen
+            = pct * gross_load - rooftop_gen
+
```
trevorb1, line 288: The implementation looks good to me (although I haven't tested it)! Im just wondering if this should be a flag option from the config? Like have we checked the different demand sources (EFS, AEO, etc) to ensure rooftop solar isn't accounted for in their projections? If we have, we should ensure that is clear. to the user. Else, just changing this to an optional flag would be good, I think? Or I may just be misunderstanding the implementation! 😅

trevorb1 (commented):
This looks cool, @WilsonHM18! Thanks for the contribution! Also, thanks for adding tests with your contribution! 🎉  Just a couple comments below

workflow/scripts/retrieve_small_scale_solar.py
WilsonHM18: Addressed. The hardcoded list is removed, state filtering is now in _SmallScaleSolarData.format_data() using a 2-character string length check consistent with _ElectricPowerOperationalData

workflow/scripts/retrieve_small_scale_solar.py
WilsonHM18: Addressed. Added SmallScaleSolar / _SmallScaleSolarData to eia.py following the EiaData/DataExtractor factory pattern. The year range (start_year, end_year) is fetched in a single API call using build_url, avoiding multiple instantiations.

workflow/scripts/opts/policy.py
WilsonHM18: Thanks for flagging this. I just added a docstring note clarifying the assumption. Do you know whether EFS and AEO demand profiles in pypsa-usa represent net or gross load? My understanding is that they're all calibrated to metered grid data (net of BTM), which would make the credit appropriate across all demand sources. Happy to add a config flag if you think that's safer given the uncertainty.

PR #774, opened 2026-08-18
Population-based demand allocation (2020 census) behind bus_allocation toggle

ktehranchi:
## What

Replaces the 2016-vintage Breakthrough Energy `Pd` column as the bus-level demand-allocation weight with **2020 Decennial Census county populations** (`data/population/DECENNIALDHC2020.P1-Data.csv`, already retrieved via the sector databundle). Each county's population is split evenly across its substations, then across each substation's buses.

- Canonical weight is the new **`n.buses.load_weight`** column; `Pd` is kept untouched for reference.
- Config toggle **`electricity.demand.bus_allocation`**: `population` (new default) | `breakthrough` (legacy). The equivalence config pins `breakthrough` (the anchor has no population method).
- Consumers switched to `load_weight`: `LAF_state`, `WritePopulation` zone→bus disaggregation, `WriteIndustrial` load-bus filter, clustering `population` weighting, substation aggregation.
- Note: PUDL was evaluated first per the original ask, but PUDL only distributes **2010** DP1 population (verified against the v2026.1.0 data dictionary); the in-repo 2020 county file is the newer federal source.

## A/B testing

New harness mode: `uv run python -m tests.equivalence.ab [--skip-solve]` builds the CA-scale pipeline twice (run names `ab_pd`/`ab_pop`) and emits `workflow/results/ab/ab_report.html`.

Results (CA, western, c4, REM-3h, solved):

| metric | legacy Pd | population | delta |
|---|---|---|---|
| load-bearing substations | 1,674 | 1,910 | +236 |
| mean total demand (MW) | 30,301 | 30,312 | +0.036% |
| mean abs per-substation shift | — | — | 7.7 MW |
| objective | -2,550,837,833 | -2,552,735,866 | 0.074% |
| capacity mix | — | — | only OCGT moves (809→727 MW) |

**Invariants (PASS):** substation topology identical; per-state demand totals scale exactly with the state's LAF share (allocation plumbing conserves demand — only the weight source moves it).

**Legacy invariance:** with `bus_allocation: breakthrough`, artifacts are **bit-identical** to pre-change equivalence artifacts (max abs diff 0.0 on `Load_t.p_set`, `Pd`, generator `p_nom`).

## Tests

- 7 new unit tests (`test_bus_population.py`); full unit suite 35 passed / 10 pre-existing skips; static tier 73 passed (incl. DAG dry-runs).
- `uv.lock` gains the lock section for the existing `test` extra (from `uv sync --extra test`); no pin changes.


ktehranchi:
Opened against the wrong repo — this belongs on the ktehranchi fork's v1-epic stack. Closing.

